PR #6, opened 2020-03-03
Miscellaneous improvements to code readability.

jonathanknowles:
## Related Issue

## Summary

- [x] Makes documentation compatible with Haddock syntax, where applicable.
- [x] Corrects grammar errors in error constructors.
- [x] Clarifies the documentation of error value constructors for `ErrCoinSelection`.
- [x] Fixes a few small other grammatical oversights in documentation.
- [x] Restructures function `atLeast` for added clarity:
    - [x] Uses pattern matching on inputs, removing the need for partial functions.
    - [x] Adds comments to clarify the purpose of each case within the `coverTarget`.
    - [x] Uses clearer names for accumulators to reflect their purpose.

### DIFF
--- a/src/Cardano/CoinSelection.hs
+++ b/src/Cardano/CoinSelection.hs
@@ -6,12 +6,9 @@
 -- Copyright: © 2018-2020 IOHK
 -- License: Apache-2.0
 --
--- Provides the API of Coin Selection algorithm and Fee Calculation
--- This module contains the implementation of adjusting coin selection for a
--- fee.  The sender pays for the fee and additional inputs are picked randomly.
--- For more information refer to:
--- https://iohk.io/blog/self-organisation-in-coin-selection/
-
+-- Provides general functions and types relating to coin selection and fee
+-- balancing.
+--
 module Cardano.CoinSelection
     (
       -- * Coin Selection
@@ -43,17 +40,21 @@ import GHC.Generics
 
 data CoinSelection = CoinSelection
     { inputs  :: [(TxIn, TxOut)]
-      -- ^ Picked inputs
+      -- ^ Picked inputs.
     , outputs :: [TxOut]
-      -- ^ Picked outputs
+      -- ^ Picked outputs.
     , change  :: [Coin]
-      -- ^ Resulting changes
+      -- ^ Resulting change.
     } deriving (Generic, Show, Eq)
 
--- NOTE
+-- NOTE:
+--
 -- We don't check for duplicates when combining selections because we assume
--- they are constructed from independent elements. In practice, we could nub
--- the list or use a `Set` ?
+-- they are constructed from independent elements.
+--
+-- As an alternative to the current implementation, we could 'nub' the list or
+-- use a 'Set'.
+--
 instance Semigroup CoinSelection where
     a <> b = CoinSelection
         { inputs = inputs a <> inputs b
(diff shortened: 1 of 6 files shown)


src/Cardano/CoinSelection.hs
```diff
@@ -6,12 +6,9 @@
 -- Copyright: © 2018-2020 IOHK
 -- License: Apache-2.0
 --
--- Provides the API of Coin Selection algorithm and Fee Calculation
--- This module contains the implementation of adjusting coin selection for a
--- fee.  The sender pays for the fee and additional inputs are picked randomly.
--- For more information refer to:
--- https://iohk.io/blog/self-organisation-in-coin-selection/
```
jonathanknowles: As far as I can tell, this module doesn't actually include the implementation, and only includes the common API, so removing this comment!

src/Cardano/Types.hs
```diff
@@ -250,10 +250,10 @@ restrictedTo (UTxO utxo) outs =
 -------------------------------------------------------------------------------}
 
 -- | Allows us to define the "domain" of any type — @UTxO@ in particular — and
--- use 'dom' to refer to the /inputs/ of an /utxo/.
+-- use 'dom' to refer to the /inputs/ of an /UTxO/.
 --
--- This is the terminology used in the [Formal Specification for a Cardano](https://github.com/input-output-hk/cardano-wallet/blob/master/specifications/wallet/formal-specification-for-a-cardano-wallet.pdf)
--- uses.
```
jonathanknowles: There was a missing "Wallet" here!

src/Cardano/CoinSelection.hs
```diff
@@ -104,23 +107,30 @@ addCoin total c = total + (fromIntegral (getCoin c))
...
-    -- inputs was reached.
+    -- ^ The maximum number of allowed inputs was not enough to cover the total
+    -- payment amount.
+    --
     | ErrInputsDepleted
-    -- ^ When trying to construct a transaction, the available inputs are
-    -- depleted even when UTxO is properly fragmented and with enough funds to
-    -- cover payment
+    -- ^ All available inputs were depleted, even though the UTxO was
+    -- sufficiently fragmented and with enough funds to cover payment.
```
jonathanknowles: We don't need to mention construction of transactions at this stage, as we're only dealing with coin selections.

src/Cardano/CoinSelection/LargestFirst.hs
```diff
@@ -34,15 +34,15 @@ import qualified Data.List as L
...
+-- | Implements the __Largest-First__ coin selection algorithm.
 largestFirst
     :: forall m e. Monad m
     => CoinSelectionOptions e
     -> NonEmpty TxOut
     -> UTxO
     -> ExceptT (ErrCoinSelection e) m (CoinSelection, UTxO)
 largestFirst opt outs utxo = do
-    let descending = NE.toList . NE.sortBy (flip $ comparing coin)
+    let outsDescending = NE.toList $ NE.sortBy (flip $ comparing coin) outs
```
jonathanknowles: This function was applied multiple times to `outs`, but not to any other values.

Therefore, we can apply it just once and then give the result a meaningful name.

src/Cardano/CoinSelection/LargestFirst.hs
```diff
@@ -94,28 +95,34 @@ atLeast
...
   where
-    coverOutput
-        :: (Integer, [(TxIn, TxOut)])
+    coverTarget
+        :: Integer
+        -> [(TxIn, TxOut)]
         -> [(TxIn, TxOut)]
         -> Maybe ([(TxIn, TxOut)], CoinSelection)
-    coverOutput (target, ins) utxo
+    coverTarget target utxoRemaining utxoSelected
```
paweljakubas: assuming the below refactoring does not destroy anything;-)

jonathanknowles: All the tests still pass!

PR #7, opened 2020-03-05
Improve documentation for largest-first algorithm.

jonathanknowles:
## Related Issue

## Summary

This goal of this PR is to improve the documentation of the largest-first algorithm and the readability of the implementation.

### Documentation Improvements

- [x] Provides Haddock documentation for the `largestFirst` function.
- [x] Revises existing the Haddock documentation for the `atLeast` function.
- [x] Revises existing the Haddock documentation for the `ErrCoinSelection` type.
- [x] Revises descriptions of test fixtures, so that terminology used is consistent with the documentation.

### Related Code Improvements

- [x] Adjusts the names of constructors for `ErrCoinSelection`, so that terminology used is consistent with the documentation.
- [x] Revises function `largestFirst` to:
    - [x] Use a single `where` clause to define all constants rather than a sequence of monadic `let` bindings.
    - [x] Use a pure constant with pattern guard to define the error condition rather than a sequence of monadic `when` clauses.
    - [x] Use names `amountRequested` and `amountAvailable`, making it obvious that these values are _directly comparable_.
    - [x] Use names `outputsDescending` and `utxoDescending`, making it obvious that these values refer to lists in _descending order_.

### DIFF
--- a/src/Cardano/CoinSelection/LargestFirst.hs
+++ b/src/Cardano/CoinSelection/LargestFirst.hs
@@ -66,7 +66,7 @@ import qualified Data.Map.Strict as Map
 -- For /each output/ in the (descending) output list, the algorithm repeatedly
 -- selects unspent values from the /remaining UTxO set/ (in descending order)
 -- until the /total selected value/ is greater than (or equal to) the output
--- value, at which point the algorithm move ons to processing the /next/
+-- value, at which point the algorithm moves on to processing the /next/
 -- output.
 --
 -- If the /total selected value/ is greater than required for a particular
--- a/src/Cardano/CoinSelection/LargestFirst.hs
+++ b/src/Cardano/CoinSelection/LargestFirst.hs
@@ -37,53 +37,109 @@ import qualified Data.Map.Strict as Map
 
 -- | Generate a coin selection according to the __largest first__ algorithm.
 --
--- For the given /output list/ and /starting UTxO set/, this algorithm
--- generates a /coin selection/ that is capable of paying for all of the
--- outputs, and a /remaining UTxO set/ from which spent values have been
--- removed.
+-- === Summary
+--
+-- For the given /output list/ and /initial UTxO set/, this algorithm generates
+-- a /coin selection/ that is capable of paying for all of the outputs, and a
+-- /remaining UTxO set/ from which spent values have been removed.
+--
+-- === State Maintained by the Algorithm
+--
+-- At all stages of processing, the algorithm maintains:
+--
+--  1.  A __/remaining UTxO list/__
+--
+--      This is initially equal to the given /initial UTxO set/ parameter,
+--      sorted into /descending order of coin value/.
+--
+--      The /head/ of the list is always the remaining UTxO entry with the
+--      /greatest coin value/.
+--
+--      Entries are incrementally removed from the /head/ of the list as the
+--      algorithm proceeds, until the list is empty.
+--
+--  2.  An __/unpaid output list/__
+--
+--      This is initially equal to the given /output list/ parameter, sorted
+--      into /descending order of coin value/.
+--
+--      The /head/ of the list is always the unpaid output with the
+--      /greatest coin value/.
+--
+--      Entries are incrementally removed from the /head/ of the list as the
+--      algorithm proceeds, until the list is empty.
+--
+--  3.  An __/accumulated coin selection/__
+--
+--      This is initially /empty/.
+--
+--      Entries are incrementally added as each output is paid for, until the
+--      /unpaid output list/ is empty.
 --
 -- === Cardinality Rules
 --
 -- The algorithm requires that:
 --
---  1.  Each output is paid for by /one or more/ entries from the UTxO set.
+--  1.  Each output from the given /output list/ is paid for by /one or more/
+--      entries from the /initial UTxO set/.
 --
---  2.  Each entry from the UTxO set is used to pay for /at most one/ output.
+--  2.  Each entry from the /initial UTxO set/ is used to pay for /at most one/
+--      output from the given /output list/.
 --
---      (A given entry from the UTxO set /cannot/ be used to pay for multiple
---      outputs.)
+--      (A single UTxO entry __cannot__ be used to pay for multiple outputs.)
 --
 -- === Order of Processing
 --
--- The algorithm processes /both/ the given output list /and/ the supplied UTxO
--- set in __descending order of coin value__, from largest to smallest.
+-- The algorithm proceeds according to the following sequence of steps:
 --
--- At all stages of processing, the algorithm maintains a /remaining UTxO set/
--- that is steadily depleted as outputs are paid for.
+--  *   /Step 1/
 --
--- === Processing an Output
+--      Remove a single /unpaid output/ from the head of the
+--      /unpaid output list/.
 --
--- For /each output/ in the (descending) output list, the algorithm repeatedly
--- selects unspent values from the /remaining UTxO set/ (in descending order)
--- until the /total selected value/ is greater than (or equal to) the output
--- value, at which point the algorithm moves on to processing the /next/
--- output.
+--  *   /Step 2/
 --
--- If the /total selected value/ is greater than required for a particular
--- output, the algorithm generates a /change output/ with the exact difference
--- in value.
+--      Repeatedly remove entries from the head of the /remaining UTxO list/
+--      until the total value of entries removed is /greater than or equal to/
+--      the value of the /unpaid output/.
 --
--- === Termination Conditions
+--  *   /Step 3/
+--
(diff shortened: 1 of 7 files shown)


src/Cardano/CoinSelection/LargestFirst.hs
```diff
@@ -20,77 +21,140 @@ import Cardano.Types
...
 import qualified Data.List as L
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
 
--- | Implements the __Largest-First__ coin selection algorithm.
+-- | Generate a coin selection according to the __largest first__ algorithm.
+--
+-- For the given /output list/ and /starting UTxO set/, this algorithm
+-- generates a /coin selection/ that is capable of paying for all of the
+-- outputs, and a /remaining UTxO set/ from which spent values have been
```
paweljakubas: /remaining UTxO set/ is the one that is the result of /starting UTxO set/ - spent values, right?

src/Cardano/CoinSelection/LargestFirst.hs
```diff
@@ -20,77 +21,140 @@ import Cardano.Types
...
+--
+-- At all stages of processing, the algorithm maintains a /remaining UTxO set/
+-- that is steadily depleted as outputs are paid for.
+--
+-- === Processing an Output
+--
+-- For /each output/ in the (descending) output list, the algorithm repeatedly
+-- selects unspent values from the /remaining UTxO set/ (in descending order)
+-- until the /total selected value/ is greater than (or equal to) the output
+-- value, at which point the algorithm move ons to processing the /next/
```
paweljakubas: not `moves on`?

src/Cardano/CoinSelection/LargestFirst.hs
```diff
@@ -20,77 +21,140 @@ import Cardano.Types
...
+--  2.  The /number/ of entries in the starting UTxO set is /smaller than/ the
+--      number of requested outputs.
+--
+--      Due to the nature of the algorithm, /at least one/ UTxO entry is
+--      required /for each/ output.
+--
+--      See: __'ErrUtxoNotFragmentedEnough'__.
+--
+--  3.  Due to the particular /distribution/ of values within the starting UTxO
+--      set, the algorithm depletes all entries from the set /before/ it is able
```
paweljakubas: maybe you meant : `and it is unable to cover all requested outputs` ?

src/Cardano/CoinSelection/LargestFirst.hs
jonathanknowles: > /remaining UTxO set/ is the one that is the result of /starting UTxO set/ - spent values, right?

Yes, that was my intention.

Do you think there is anything I can add or change to make it clearer?

src/Cardano/CoinSelection/LargestFirst.hs
```diff
@@ -20,77 +21,140 @@ import Cardano.Types
...
+-- until the /total selected value/ is greater than (or equal to) the output
+-- value, at which point the algorithm move ons to processing the /next/
+-- output.
+--
+-- If the /total selected value/ is greater than required for a particular
+-- output, the algorithm generates a /change output/ with the exact difference
+-- in value.
+--
+-- === Termination Conditions
+--
```
jonathanknowles: Perhaps it would be worth adding the conditions under which successful termination occurs here.

src/Cardano/CoinSelection/LargestFirst.hs
jonathanknowles: Thanks for spotting this! Fixed in https://github.com/input-output-hk/cardano-coin-selection/pull/7/commits/6c8c3ce8fe11f2a492e43a8fe073ca48c46f4476.

src/Cardano/CoinSelection/LargestFirst.hs
jonathanknowles: In this case, I think the wording makes sense.

I tend to view this algorithm as a **race** between:

- Depleting the remaining UTxO
- Depleting the remaining list of unpaid outputs

We **ideally** want to deplete the list of unpaid outputs **before** we deplete the remaining UTxO.

The failure case occurs when we deplete the remaining UTxO before we are able to deplete the list of unpaid outputs, i.e., _before we are able to pay for all the outputs_.

I do think the current phrasing is correct, but I'll have another think about whether there's a better way to phrase this.

src/Cardano/CoinSelection/LargestFirst.hs
jonathanknowles: Fixed. See: https://github.com/input-output-hk/cardano-coin-selection/pull/7/commits/973126e81272aaef88f4875f7d6a75c5956ea95e

jonathanknowles:
> very nice description of algorithm!

Thanks! I hope that it will also seem clear to people that have never read the code before. In the future, we can also ask people unfamiliar with these algorithms to give feedback.

> I hope the code refactoring is ok not only from the point of view of unit tests but also some regression tests we have in `cardano-wallet` repo.

That is a very good point. However, we should still be covered by this integration test when `cardano-wallet` eventually becomes a consumer of the `cardano-coin-selection` library. (I'm assuming we will leave that integration test in place.)

src/Cardano/CoinSelection/LargestFirst.hs
jonathanknowles: @paweljakubas I've updated the phrasing with this change: 

https://github.com/input-output-hk/cardano-coin-selection/pull/7/commits/973126e81272aaef88f4875f7d6a75c5956ea95e

src/Cardano/CoinSelection/LargestFirst.hs
```diff
@@ -20,77 +21,196 @@ import Cardano.Types
...
 import qualified Data.Map.Strict as Map
 
--- | Implements the __Largest-First__ coin selection algorithm.
+-- | Generate a coin selection according to the __largest first__ algorithm.
+--
+-- === Summary
+--
+-- For the given /output list/ and /initial UTxO set/, this algorithm generates
+-- a /coin selection/ that is capable of paying for all of the outputs, and a
+-- /remaining UTxO set/ from which spent values have been removed.
```
jonathanknowles:
```suggestion
-- /remaining UTxO set/ from which all spent values have been removed.
```

src/Cardano/CoinSelection/LargestFirst.hs
```diff
@@ -20,77 +21,196 @@ import Cardano.Types
...
+--  2.  The /number/ of entries in the initial UTxO set is /smaller than/ the
+--      number of requested outputs.
+--
+--      Due to the nature of the algorithm, /at least one/ UTxO entry is
+--      required /for each/ output.
+--
+--      See: __'ErrUtxoNotFragmentedEnough'__.
+--
+--  3.  Due to the particular /distribution/ of values within the initial UTxO
+--      set, the algorithm depletes all entries from the set /before/ it is
```
jonathanknowles:
```suggestion
--      set, the algorithm depletes all entries from the UTxO set /before/ it is
```

src/Cardano/CoinSelection/LargestFirst.hs
```diff
@@ -20,77 +21,196 @@ import Cardano.Types
...
+--  *   /Step 2/
+--
+--      Repeatedly remove entries from the head of the /remaining UTxO list/
+--      until the total value of entries removed is /greater than or equal to/
+--      the value of the /unpaid output/.
+--
+--  *   /Step 3/
+--
+--      Use the /removed UTxO entries/ to pay for the /unpaid output/, by
+--      adding entries to the /accumulated coin selection/.
```
jonathanknowles: What kind of entries are added to the accumulated coin selection?

We should make this clearer.

src/Cardano/CoinSelection/LargestFirst.hs
jonathanknowles: Fixed in https://github.com/input-output-hk/cardano-coin-selection/pull/7/commits/aae8f10ba8cd883dbed18881376d293b47c6133e.

src/Cardano/CoinSelection/LargestFirst.hs
jonathanknowles: Fixed in https://github.com/input-output-hk/cardano-coin-selection/pull/7/commits/4f4e71bdfc57ddab0666b4dbb1d29b285fb0641f.

src/Cardano/CoinSelection/LargestFirst.hs
jonathanknowles: Fixed in https://github.com/input-output-hk/cardano-coin-selection/pull/7/commits/49339a3dbc7c0bef65106d9a909a3a8b7d0b36ab.

PR #9, opened 2020-03-06
Add documentation to the `CoinSelection` module.

jonathanknowles:
## Related Issue

## Summary

This PR:

- [x] adds Haddock documentation to functions and types within the `CoinSelection` module.
- [x] revises the grouping of exported types into logical sections.
- [x] moves utility functions (not part of the public API) to the end of the module.

### DIFF
--- a/src/Cardano/CoinSelection.hs
+++ b/src/Cardano/CoinSelection.hs
@@ -11,14 +11,17 @@
 --
 module Cardano.CoinSelection
     (
-      -- * Coin Selection
-      CoinSelection(..)
+      -- * Types
+      CoinSelection (..)
+    , CoinSelectionOptions (..)
+    , ErrCoinSelection (..)
+
+      -- * Calculating Balances
     , inputBalance
     , outputBalance
     , changeBalance
     , feeBalance
-    , ErrCoinSelection (..)
-    , CoinSelectionOptions (..)
+
     ) where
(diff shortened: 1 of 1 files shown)


src/Cardano/CoinSelection.hs
```diff
@@ -57,22 +68,24 @@ data CoinSelection = CoinSelection
...
 
 instance Buildable CoinSelection where
-    build (CoinSelection inps outs chngs) = mempty
-        <> nameF "inputs" (blockListF' "-" inpsF inps)
-        <> nameF "outputs" (blockListF outs)
-        <> nameF "change" (listF chngs)
+    build cs = mempty
+        <> nameF "inputs"  (blockListF' "-" inpsF $ inputs  cs)
+        <> nameF "outputs" (blockListF            $ outputs cs)
+        <> nameF "change"  (listF                 $ change  cs)
```
paweljakubas: this is compliant with our coding standards?

jonathanknowles: @paweljakubas wrote:

> this is compliant with our coding standards?

You're right -- it is not compliant.

I was under the impression that we had recently relaxed this standard, but it turns out that we haven't!

Therefore, I have removed this change (and replaced it with a simplification).

Thanks for pointing this out!

PR #10, opened 2020-03-09
Introduce an abstract `CoinSelectionAlgorithm` type.

jonathanknowles:
## Related Issue

https://github.com/input-output-hk/cardano-wallet/issues/1381

## Summary

This PR introduces an abstract `CoinSelectionAlgorithm` type.

This type provides a _shared interface_ that all coin selection algorithms implement, as well as a place to document the common features of such algorithms.

## Limitations

The `CoinSelectionAlgorithm` abstraction does not deal with migration, as this is a rather special case.

### DIFF
--- a/src/Cardano/CoinSelection/LargestFirst.hs
+++ b/src/Cardano/CoinSelection/LargestFirst.hs
@@ -1,5 +1,3 @@
-{-# LANGUAGE RankNTypes #-}
-
 -- |
 -- Copyright: © 2018-2020 IOHK
 -- License: Apache-2.0
@@ -162,11 +160,11 @@ import qualified Data.Map.Strict as Map
 --
 --      See: __'ErrMaximumInputCountExceeded'__.
 --
-largestFirst :: forall m e. Monad m => CoinSelectionAlgorithm m e
+largestFirst :: Monad m => CoinSelectionAlgorithm m e
 largestFirst = CoinSelectionAlgorithm payForOutputs
 
 payForOutputs
-    :: forall m e. Monad m
+    :: Monad m
     => CoinSelectionOptions e
     -> NonEmpty TxOut
     -> UTxO
--- a/src/Cardano/CoinSelection/LargestFirst.hs
+++ b/src/Cardano/CoinSelection/LargestFirst.hs
@@ -9,13 +9,16 @@
 --
 module Cardano.CoinSelection.LargestFirst (
     largestFirst
-  , atLeast
   ) where
 
 import Prelude
 
 import Cardano.CoinSelection
-    ( CoinSelection (..), CoinSelectionOptions (..), ErrCoinSelection (..) )
+    ( CoinSelection (..)
+    , CoinSelectionAlgorithm (..)
+    , CoinSelectionOptions (..)
+    , ErrCoinSelection (..)
+    )
 import Cardano.Types
     ( Coin (..), TxIn, TxOut (..), UTxO (..), balance )
 import Control.Arrow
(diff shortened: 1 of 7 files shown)


test/Cardano/CoinSelection/LargestFirstSpec.hs
```diff
@@ -229,8 +233,10 @@ propDeterministic
     -> Property
 propDeterministic (CoinSelProp utxo txOuts) = do
     let opts = CoinSelectionOptions (const 100) noValidation
-    let resultOne = runIdentity $ runExceptT $ largestFirst opts txOuts utxo
-    let resultTwo = runIdentity $ runExceptT $ largestFirst opts txOuts utxo
+    let resultOne = runIdentity $ runExceptT $
+            selectCoins largestFirst opts txOuts utxo
+    let resultTwo = runIdentity $ runExceptT $
+            selectCoins largestFirst opts txOuts utxo
     resultOne === resultTwo
```
jonathanknowles: I suspect that this property is actually redundant, since `largestFirst` is a pure function.

src/Cardano/CoinSelection/LargestFirst.hs
```diff
@@ -37,12 +38,6 @@ import qualified Data.Map.Strict as Map
 
 -- | Generate a coin selection according to the __largest first__ algorithm.
 --
--- === Summary
---
--- For the given /output list/ and /initial UTxO set/, this algorithm generates
--- a /coin selection/ that is capable of paying for all of the outputs, and a
--- /remaining UTxO set/ from which all spent values have been removed.
---
```
jonathanknowles: This section is no longer necessary, as we have equivalent documentation in the `CoinSelectionAlgorithm` type.

test/Cardano/CoinSelectionSpec.hs
```diff
@@ -146,21 +150,17 @@ data CoinSelectionResult = CoinSelectionResult
...
-    :: ( CoinSelectionOptions ErrValidation
-         -> NonEmpty TxOut
-         -> UTxO
-         -> ExceptT (ErrCoinSelection ErrValidation) IO (CoinSelection, UTxO)
-       )
+    :: CoinSelectionAlgorithm IO ErrValidation
     -> String
     -> Either (ErrCoinSelection ErrValidation) CoinSelectionResult
     -> CoinSelectionFixture
     -> SpecWith ()
```
jonathanknowles: Using the `CoinSelectionAlgorithm` type here makes it clear that this function will work on any coin selection algorithm.

test/Cardano/CoinSelection/LargestFirstSpec.hs
jonathanknowles: This property has since been modified to be useful again.

src/Cardano/CoinSelection/Random.hs
```diff
@@ -106,35 +109,38 @@ data TargetRange = TargetRange
 -- that a randomly chosen UTxO entry will push the total above the upper bound
 -- we set.
 -- @
-random
-    :: forall m e. MonadRandom m
+randomImprove :: MonadRandom m => CoinSelectionAlgorithm m e
+randomImprove = CoinSelectionAlgorithm payForOutputs
```
paweljakubas: `randomImprove` immediately associates t me with `improve` below. but that is minor

jonathanknowles: My motivation here was to make the names of identifier match up with the names of algorithms. So we have:

1. "Largest-First" ↔ `largestFirst`.
2. "Random-Improve" ↔ `randomImprove`.

jonathanknowles:
After a discussion within our team, we have decided to:

1. Merge the `CoinSelectionAlgorithm` abstraction in this PR, as it provides a definite improvement over the status quo. It codifies an implied abstraction that already exists in our code in several places (including the test suite), and provides a common place for documentation.

2. At some point later on, investigate how _migration_ can be included in the abstraction. We could achieve this by creating a separate abstraction, or we could find a way to decompose `Migration.depleteUTxO` so that migration can be included in the abstraction.

PR #12, opened 2020-03-11
Add licence badge

paweljakubas:
small thing worth adding to README.md


jonathanknowles:
Looks good!

The only thing (and its a minor thing):

Perhaps we could move it to the top, next to the other badge?

PR #13, opened 2020-03-12
Revise the `Random-Improve` module.

jonathanknowles:
## Related Issue

## Summary

The primary purpose of this PR is to revise documentation for the `Random-Improve` module.

However, this PR also makes a small number of _related_ changes:

- [x] Adjusts documentation for `Largest-First` so that both coin selection algorithms have a similar documentation structure.
- [x] Introduces the type synonym `CoinSelectionInput = (TxIn, TxOut)` to aid readability, and to make it explicit that values of this type are intended to be inputs to a `CoinSelection`.
- [x] Adjusts the naming of some identifiers for consistency across modules and to be consistent with terminology used in the documentation. (So that people reading the code are presented with a consistent vocabulary.)
- [x] Revises the definitions of some functions so that constants are defined within `where` clauses instead of within monadic `let` bindings.
- [x] Groups related utility types and functions together at the end of the module.

### DIFF
--- a/src/Cardano/CoinSelection/LargestFirst.hs
+++ b/src/Cardano/CoinSelection/LargestFirst.hs
@@ -2,7 +2,7 @@
 -- Copyright: © 2018-2020 IOHK
 -- License: Apache-2.0
 --
--- This module contains an implementation of the __largest first__ coin
+-- This module contains an implementation of the __Largest-First__ coin
 -- selection algorithm.
 --
 module Cardano.CoinSelection.LargestFirst (
@@ -36,7 +36,7 @@ import qualified Data.List as L
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Map.Strict as Map
 
--- | Generate a coin selection according to the __largest first__ algorithm.
+-- | An implementation of the __Largest-First__ coin selection algorithm.
 --
 -- === State Maintained by the Algorithm
 --
--- a/src/Cardano/CoinSelection/LargestFirst.hs
+++ b/src/Cardano/CoinSelection/LargestFirst.hs
@@ -38,7 +38,16 @@ import qualified Data.Map.Strict as Map
 
 -- | An implementation of the __Largest-First__ coin selection algorithm.
 --
--- === State Maintained by the Algorithm
+-- = Overview
+--
+-- The __Largest-First__ algorithm processes outputs in /descending order of/
+-- /value/, from /largest/ to /smallest/.
+--
+-- For each output, it repeatedly selects the /largest/ remaining unspent UTxO
+-- entry until the value of selected entries is greater than or equal to the
+-- value of that output.
+--
+-- = State Maintained by the Algorithm
 --
 -- At all stages of processing, the algorithm maintains:
 --
@@ -48,7 +57,7 @@ import qualified Data.Map.Strict as Map
 --      sorted into /descending order of coin value/.
 --
 --      The /head/ of the list is always the remaining UTxO entry with the
---      /greatest coin value/.
+--      /largest coin value/.
 --
 --      Entries are incrementally removed from the /head/ of the list as the
 --      algorithm proceeds, until the list is empty.
@@ -59,7 +68,7 @@ import qualified Data.Map.Strict as Map
 --      into /descending order of coin value/.
 --
 --      The /head/ of the list is always the unpaid output with the
---      /greatest coin value/.
+--      /largest coin value/.
 --
 --      Entries are incrementally removed from the /head/ of the list as the
 --      algorithm proceeds, until the list is empty.
@@ -71,7 +80,7 @@ import qualified Data.Map.Strict as Map
 --      Entries are incrementally added as each output is paid for, until the
 --      /unpaid output list/ is empty.
 --
--- === Cardinality Rules
+-- = Cardinality Rules
 --
 -- The algorithm requires that:
 --
@@ -83,7 +92,7 @@ import qualified Data.Map.Strict as Map
 --
 --      (A single UTxO entry __cannot__ be used to pay for multiple outputs.)
 --
--- === Order of Processing
+-- = Order of Processing
 --
 -- The algorithm proceeds according to the following sequence of steps:
 --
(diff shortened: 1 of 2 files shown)


src/Cardano/CoinSelection/LargestFirst.hs
```diff
@@ -48,7 +57,7 @@ import qualified Data.Map.Strict as Map
 --      sorted into /descending order of coin value/.
 --
 --      The /head/ of the list is always the remaining UTxO entry with the
---      /greatest coin value/.
+--      /largest coin value/.
```
jonathanknowles: This might seem like a trivial change, but the wording is now consistent with the name of the algorithm.

src/Cardano/CoinSelection/LargestFirst.hs
```diff
@@ -59,7 +68,7 @@ import qualified Data.Map.Strict as Map
 --      into /descending order of coin value/.
 --
 --      The /head/ of the list is always the unpaid output with the
---      /greatest coin value/.
+--      /largest coin value/.
```
jonathanknowles: This might seem like a trivial change, but the wording is now consistent with the name of the algorithm.

src/Cardano/CoinSelection/Random.hs
```diff
@@ -49,66 +49,168 @@ import Data.Functor
...
+-- For each output of value __/v/__, the algorithm /randomly/ selects entries
+-- from the /remaining UTxO set/, until the total value of selected entries is
+-- greater than or equal to __/v/__. The selected entries are then associated
+-- with that output, and removed from the /remaining UTxO set/.
+--
+-- This phase ends when every output has been associated with a selection of
+-- UTxO entries.
+--
+-- However, if the remaining UTxO set is completely exhausted before all
+-- outputs can be processed, the algorithm terminates and delegates to the
```
paweljakubas: maybe it is better to say instead of delegates `falls back to` (just hunch;-)

src/Cardano/CoinSelection/Random.hs
```diff
@@ -49,66 +49,168 @@ import Data.Functor
...
+-- For each output of value __/v/__, the algorithm:
+--
+--  1.  __Calculates a /target range/__ for the total value of inputs used to
+--      pay for that output, defined by the triplet:
+--
+--      (/minimum/, /ideal/, /maximum/) = (/v/, /2v/, /3v/)
+--
+--  2.  __Attempts to /improve/ upon the /existing UTxO selection/__ for that
+--      output, by repeatedly selecting additional entries at random from the
+--      /remaining UTxO set/, stopping when the selection can be improved upon
```
paweljakubas: you mean `cannot be further improved`, right?

src/Cardano/CoinSelection/Random.hs
```diff
@@ -49,66 +49,168 @@ import Data.Functor
...
+--
+-- There are several motivating principles behind the design of the algorithm.
+--
+-- === Principle 1: Dust Management
+--
+-- The probability that random selection will choose dust entries from a UTxO
+-- set increases with the proportion of dust in the set.
+--
+-- Therefore, for a UTxO set with a large amount of dust, there's a high
+-- probability that a random subset will include a large amount of dust.
```
paweljakubas: the trade-off here, the tx will cost more as tx fee is bigger, but it has also benefits

src/Cardano/CoinSelection/Random.hs
```diff
@@ -118,55 +220,72 @@ payForOutputs
...
+        Just (inputCountRemaining, utxoRemaining, randomSelections) -> do
+            (_, finalSelection, utxoRemaining') <- lift $ foldM
+                improveSelection
+                    (inputCountRemaining, mempty, utxoRemaining)
+                    (reverse randomSelections)
+            validateSelection finalSelection $>
+                (finalSelection, utxoRemaining')
         Nothing ->
-            selectCoins largestFirst opt outs utxo
+            -- In the case that we fail to generate a selection, delegate to
```
paweljakubas: once again, if falls back is better expression here maybe we can improve here also;-)

src/Cardano/CoinSelection/Random.hs
```diff
@@ -118,55 +220,72 @@ payForOutputs
...
+        L.sortOn (Down . coin) $ NE.toList outputsRequested
+    validateSelection =
+        except . left ErrInvalidSelection . validate options
 
--- | Perform a random selection on a given output, without improvement.
-makeSelection
+-- | Randomly select entries from the given UTxO set, until the total value of
+--   selected entries is greater than or equal to the given output value.
+--
+-- Once a random selection has been made that meets the above criterion, this
```
paweljakubas: `Once a random selection  that meets the above criterion has been made ` sounds better for me, but you are native speaker here :-)

jonathanknowles: This one is okay! Both phrases are equivalent. :)

src/Cardano/CoinSelection/Random.hs
jonathanknowles: Yes, that's precisely the meaning!

This wording is actually okay.

We could say "cannot be further improved upon", but in this context, the original wording seems just as natural.

The "can be _verb_-ed no further" phrase is similar to "can _verb_ no further".

Similar examples:

> "We stopped walking when our legs could carry us no further."
> "After we reached an obstacle in the road, we could drive no further."
> "The algorithm stops when the counter can be decreased no further."
> "The sorting procedure halts when the the list can be shortened no further."
> "We continue until the metal reaches absolute zero and can be cooled no further."

src/Cardano/CoinSelection/Random.hs
jonathanknowles: Fair enough!

Fixed in 60258d5fbc5bc272b41d8f68df358a5ad75c47da

src/Cardano/CoinSelection/Random.hs
jonathanknowles: Fair enough!

Fixed in 60258d5fbc5bc272b41d8f68df358a5ad75c47da.

jonathanknowles:
@paweljakubas wrote:
> very nice description of how random algo works, meticulously formatted and executed 100

Thanks for taking the time to read through and check it!

I think one more thing we could do (in future) is to ask someone unfamiliar with the algorithms to read through these descriptions and see if they make sense.

When we write the human-readable specification, it will be especially useful to go through such an exercise.

PR #16, opened 2020-03-20
Coin Selection Specification: Draft 1

jonathanknowles:
## Related Issue

https://github.com/input-output-hk/cardano-coin-selection/issues/21

## Summary

This PR contains a draft of a specification for the coin selection algorithms used by Cardano Wallet.

The purpose of the specification is to provide an easy-to-understand, **_human-readable_** description of these algorithms that doesn't require an understanding of Haskell.

The purpose of this PR is to gain feedback on areas for improvement. Please feel free to leave comments and constructive suggestions!

The document is written in Markdown, and can be read in its [rendered form](https://github.com/input-output-hk/cardano-coin-selection/blob/jonathanknowles/specification-draft-1/information/specification.md).

Please note that there are several **areas not addressed by this specification**:

* There is no description of the algorithm used for **migration** of funds from one wallet to another.
* Descriptions of **fee balancing** algorithms are not included.


information/specification.md
```diff
@@ -0,0 +1,705 @@
...
+# Definitions
+
+### Address
+
+An _address_ is a unique identifier that represents a destination for a payment.
+
+Addresses are owned (and generated) by individual wallets.
+
+Coin selection algorithms are, in general, agnostic to the details of
+individual address formats. Therefore, such formats will not be discussed here.
```
jonathanknowles:
```suggestion
In general, coin selection algorithms are agnostic to the choice of address
format. Therefore, the details of address formats are not described within this
document.
```

information/specification.md
```diff
@@ -0,0 +1,705 @@
...
+## Failure Modes
+
+There are a number of ways in which a coin selection algorithm can fail:
+
+  * #### UTxO Balance Insufficient
+
+    This failure occurs when the total value of the entries within the [initial
+    UTxO set](#initial-utxo-set) (the amount of money _available_) is _less
+    than_ the the total value of all entries in the [initial output
+    list](#requested-output-list) (the amount of money _required_).
```
jonathanknowles:
```suggestion
    than_ the the total value of all entries in the [requested output
    list](#requested-output-list) (the amount of money _required_).
```

information/specification.md
```diff
@@ -0,0 +1,705 @@
...
+
+  * #### UTxO Not Fragmented Enough
+
+    This failure occurs when the _number_ of entries in the [initial UTxO
+    set](#initial-utxo-set) is _smaller than_ the number of entries in the
+    [requested output list](#requested-output-list).
+
+    All algorithms require that there is _at least one_ UTxO entry available
+    _for each_ output.
```
jonathanknowles:
```suggestion
    _for each_ requested output.
```

information/specification.md
```diff
@@ -0,0 +1,705 @@
...
+These algorithms implement a [common interface](#interface), as described
+above.
+
+There are two main algorithms used by Cardano Wallet:
+
+  * [Largest-First](#largest-first)
+  * [Random-Improve](#random-improve)
+
+In general, Cardano Wallet gives _priority_ to the
+[Random-Improve](#random-improve) algorithm.
```
jonathanknowles: Perhaps add a short paragraph here explaining *why* Cardano Wallet gives priority to Random-Improve. This paragraph could link to the "Self Organization in Coin Selection" blog article (listed at the end as an external resource).

information/specification.md
```diff
@@ -0,0 +1,705 @@
...
+selections made in the previous phase, by conservatively expanding the
+selection made for each output in order to generate improved change
+values.
+
+During this phase, the algorithm:
+
+  * processes outputs in _ascending order of coin value_.
+
+  * continues to select values from the [available UTxO
+    set](#available-utxo-set)
```
jonathanknowles:
```suggestion
    set](#available-utxo-set).
```

information/specification.md
```diff
@@ -0,0 +1,705 @@
...
+
+  * incrementally populates the
+    [accumulated coin selection](#accumulated-coin-selection-1).
+
+For each output of value **_v_**, the algorithm:
+
+ 1.  **Calculates a _target range_** for the total value of inputs used to
+     pay for that output, defined by the triplet:
+
+     (_minimum_, _ideal_, _maximum_) = (_v_, _2v_, _2v_)
```
jonathanknowles: This is a typo. The _maximum_ should be _3v_.

```suggestion
     (_minimum_, _ideal_, _maximum_) = (_v_, _2v_, _3v_)
```

information/specification.md
```diff
@@ -0,0 +1,705 @@
...
+
+        _v1_ ≤ _maximum_
+
+      * **Condition 3**: when counting cumulatively across all outputs
+        considered so far, we have not selected more than the _maximum_ number
+        of UTxO entries specified by [Maximum Input
+        Count](#maximum-input-count).
+
+ 3.  **Creates a _change value_** for the output, equal to the total value
+     of the _final UTxO selection_ for that output minus the value _v_ of
```
jonathanknowles: Consider a change in terminology here:
```suggestion
     of the _improved UTxO selection_ for that output minus the value _v_ of
```

information/specification.md
```diff
@@ -0,0 +1,706 @@
...
+  * **_o_** is a [transaction output](#transaction-output).
+
+A given [transaction input](#transaction-input) can only appear _once_ in a
+given UTxO set.
+
+### Transaction Input
+
+A _transaction input_ is a pair of values (**_h_**, **_n_**), where:
+
+  * **_h_** is the hash of an existing transaction **_t_**;
```
rvl: Any reason why this is hash and not transaction id, which is more general?

You could add a definition for Transaction ID as a unique identifier that represents a transaction. Coin selection algorithms are agnostic to the format of txid, but in cardano mainnet, it's the blake2b-256 hash of the cbor-encoded transaction.

information/specification.md
```diff
@@ -0,0 +1,706 @@
...
+A coin selection function accepts the following parameters:
+
+ 1. #### Requested Output List
+
+    A list of payments to be made to recipient addresses, encoded as a list of
+    [transaction outputs](#transaction-output).
+
+ 2. #### Initial UTxO Set
+
+    A subset of a wallet's [UTxO set](#utxo-set).
```
rvl: You could say "The wallet's UTxO set, or a subset of it." Then the last sentence won't be necessary.

information/specification.md
```diff
@@ -0,0 +1,706 @@
...
+    It represents the set of values that remain after the coin selection
+    algorithm has removed values to pay for entries in the [requested output
+    list](#requested-output-list).
+
+    If a coin selection algorithm is applied to the _complete_ UTxO set of a
+    wallet, then the remaining UTxO set represents the _updated_ UTxO set of the
+    wallet.
+
+## Properties
+
```
rvl: I wonder if this section needs revision so that it's more human readable.

Will we also have a formal specification of coin selection?

information/specification.md
```diff
@@ -0,0 +1,706 @@
...
+  * V<sub>_selected_</sub>
+
+    is the total value of the _inputs_ field of the [coin
+    selection](#coin-selection) result.
+
+### Conservation of UTxO
+
+This property states that the [initial UTxO set](#initial-utxo-set) is
+_conserved_ in the [results](#results).
```
rvl: This seems to be a rewording of "conservation of UTxO".

Maybe say something like "This property states that every UTxO in the initial set is included in either the coin selection inputs, or in the remaining utxo set, but not both."

information/specification.md
```diff
@@ -0,0 +1,706 @@
...
+  * #### UTxO Not Fragmented Enough
+
+    This failure occurs when the _number_ of entries in the [initial UTxO
+    set](#initial-utxo-set) is _smaller than_ the number of entries in the
+    [requested output list](#requested-output-list).
+
+    All algorithms require that there is _at least one_ UTxO entry available
+    _for each_ requested output.
+
+  * #### UTxO Fully Depleted
```
rvl: What is the difference between fully depleted and insufficient balance?

rvl (approved):
Nice doc - very clear.

Will there be a discussion of the tx fee being dependent on the coin selection which is dependent on tx fee? Is that what fee balancing is?

information/specification.md
```diff
@@ -0,0 +1,706 @@
...
+The purpose of this article is to describe, in _human-readable terms_, the coin
+selection algorithms used by Cardano Wallet and other parts of the Cardano
+ecosystem.
+
+# Background
+
+## What is Coin Selection?
+
+Coin selection is the process of choosing unspent coins from a wallet in order
+to pay money to one or more recipients.
```
KtorZ: Shouldn't the document actually starts with the definitions? For example, here we are already talking about _coins_ and _wallet_ without having really defined those terms. Unless the sense given to these terms is the common sense of a coin and wallet.

information/specification.md
```diff
@@ -0,0 +1,706 @@
...
+   include that dust in a given transaction, we may run out of space (by
+   reaching the transaction size limit) before we can cover the target amount.
+
+ * The most obvious strategy for paying change, which consists of making a
+   single change output with the exact excess value, will tend (over time) to
+   reduce the size of the UTxO set. This is bad for two reasons:
+
+    1. Having a small UTxO set limits the number of payments that we can make
+       in parallel. Since a single UTxO entry can only be used to pay for a
+       single output, we need at least as many UTxO entries as there are
```
KtorZ: > Since a single UTxO entry can only be used to pay for a single output

This is **not** generally true and in particular, not true for Cardano. This is a limitation our coin selection algorithm self-imposed to itself to make the problem easier to solve (so a multi-variable problem becomes multiple independant single-variable problems, easier to solve).

information/specification.md
```diff
@@ -0,0 +1,706 @@
...
+       privacy issues is beyond the scope of this article.
+
+### Goals
+
+There are several desirable properties that we would like a coin selection
+algorithm to have. These properties include:
+
+ * A coin selection algorithm should, over the course of time, aim to generate
+   and maintain a UTxO set with _useful_ outputs: that is, outputs that allow
+   us to process future payments with a _minimum_ number of inputs.
```
KtorZ: I'd use `reasonably small` instead of `minimum` here :)

information/specification.md
```diff
@@ -0,0 +1,706 @@
...
+
+ * A coin selection algorithm should employ strategies to limit the
+   amount of _dust_ that accumulates in the UTxO set.
```
KtorZ: Isn't this second property redundant with the first one?

information/specification.md
```diff
@@ -0,0 +1,706 @@
...
+For more information on dust avoidance, see [Self Organisation in Coin
+Selection](#self-organization-in-coin-selection).
+
+# Definitions
+
+### Address
+
+An _address_ is a unique identifier that represents a destination for a payment.
+
+Addresses are owned (and generated) by individual wallets.
```
KtorZ: *are typically owned 

In practice, HD wallets aren't necessary to generate addresses and interact with the blockchain.

information/specification.md
```diff
@@ -0,0 +1,706 @@
...
+
+In general, coin selection algorithms are agnostic to the choice of address
+format. Therefore, the details of individual address formats are not described
+within this document.
+
+### Coin Amount
+
+A _coin amount_ is a non-negative integer value that represents a number of
```
KtorZ: It even is a _positive_ integer value. There's in principle, no such thing as a coin of value `0` (although quite practical in many calculations). This is true at least in Byron and I know there was discussion for Shelley about whether or not a transaction with an output of value `0` should be considered valid.

information/specification.md
```diff
@@ -0,0 +1,706 @@
...
+One [Ada](https://cardanodocs.com/cardano/monetary-policy/) is _exactly_ equal
+to one million Lovelace.
+
+### UTxO Set
+
+A _UTxO set_ represents the unspent value associated with a wallet.
+
+Each member of the set is a pair of the form (**_i_**, **_o_**), where:
+
+  * **_i_** is a [transaction input](#transaction-input);
```
KtorZ: A UTxO is actually a transaction identifier and an output index. I am not sure to agree with the defnition here :)

information/specification.md
KtorZ: Agree, it doesn't quite matter what `h` is here (a hash, an encrypted digest .. whatever) but what's important though is that `h` must be a **unique** transaction identifier.

information/specification.md
KtorZ: (see also: https://github.com/input-output-hk/cardano-wallet/wiki/About-UTxO )

information/specification.md
KtorZ: Or "A UTxO set to choose from", that it comes from a wallet is quite anecdotal.

information/specification.md
```diff
@@ -0,0 +1,706 @@
...
+  * **_o_** is a [transaction output](#transaction-output).
+
+A given [transaction input](#transaction-input) can only appear _once_ in a
+given UTxO set.
+
+### Transaction Input
```
KtorZ: Note that a transaction input is **by definition**, a UTxO.

information/specification.md
```diff
@@ -0,0 +1,706 @@
...
+    The _remaining UTxO set_ is a subset of the [initial UTxO
+    set](#initial-utxo-set).
+
+    It represents the set of values that remain after the coin selection
+    algorithm has removed values to pay for entries in the [requested output
+    list](#requested-output-list).
+
+    If a coin selection algorithm is applied to the _complete_ UTxO set of a
+    wallet, then the remaining UTxO set represents the _updated_ UTxO set of the
+    wallet.
```
KtorZ: I wouldn't put too much emphasis on wallets in this document. A coin selection selects elements from a UTxO set. Period. This can obviously be applied to UTxO management inside a wallet, but that is out-of-scope of the current spec IMO.

jonathanknowles:
@rvl wrote:
> Will there be a discussion of the tx fee being dependent on the coin selection which is dependent on tx fee? Is that what fee balancing is?

Yes! There will definitely be a discussion about fee balancing.

At the moment, I'm planning to put this in the next draft. Will let you know when it's ready for review. :+1:

information/specification.md
jonathanknowles: > It even is a _positive_ integer value. There's in principle, no such thing as a coin of value `0` (although quite practical in many calculations). This is true at least in Byron and I know there was discussion for Shelley about whether or not a transaction with an output of value `0` should be considered valid.

Good catch. I initially wanted to include the possibility of zero, as this would allow us to talk about temporary amounts used in calculations (which occasionally end up as `0`, only to be filtered out later on). On reflection, I think it makes sense to exclude this case.

Fixed in 64e0dfc49e88f2ff1c3bd93a279b3a44d0d9ceec.

information/specification.md
jonathanknowles: > *are typically owned
> 
> In practice, HD wallets aren't necessary to generate addresses and interact with the blockchain.

Good point!

Fixed in 786e107d21138b5ab9797f486ac2bc96f159f405.

information/specification.md
jonathanknowles: @rvl @KtorZ 

I've attempted to address this issue here: 4ea2f9089c135796f8fdd873a2749eff1c9b3516

Hopefully the new wording is sufficiently general, but let me know if you think it could be improved further. :+1:

information/specification.md
jonathanknowles: > Maybe say something like "This property states that every UTxO in the initial set is included in either the coin selection inputs, or in the remaining utxo set, but not both."

Thanks, this wording is very clear and concise!

Fixed in e2a6a6ad2f50bc1f79b70a35b7bcfb5a82c2058b.

information/specification.md
jonathanknowles: @rvl wrote:
> I wonder if this section needs revision so that it's more human readable.

Thanks, this kind of feedback is very useful.

I'm currently attempting to write each property definition so that it has _both_ a human-readable explanation (with natural language), _and_ a more mathematical definition underneath. This way, people who are not interested in mathematics will still be able to gain an intuition of each property, but those who require more detail can read the mathematical definitions.

Do you think the human-readable explanations are accessible enough?

My interpretation of the [requirements for this specification](https://jira.iohk.io/browse/ADP-189) is that someone who's never seen the Haskell code _should_ be able to read it and then be able to re-implement these algorithms (in their language of choice). Of course, with a human language explanation, it's possible that we'll miss something. But I think it's okay for us to make further revisions to the specification if people identify things that are not clear enough.

> Will we also have a formal specification of coin selection?

I'm not sure of the answer to that! @KtorZ, are you able to shed some light on this area?

information/specification.md
jonathanknowles: @KtorZ wrote:
> I'd use `reasonably small` instead of `minimum` here :)

The "minimum number of inputs" wording actually appears in [Self Organisation in Coin Selection](https://iohk.io/en/blog/posts/2018/07/03/self-organisation-in-coin-selection/).

However, I think "reasonably small" also works well here.

Fixed in e914b068a57c76fe423545eca5c8ad315a05a068.

information/specification.md
jonathanknowles: @KtorZ wrote:
> Isn't this second property redundant with the first one?

I'm not sure that it is redundant.

In theory, it seems possible to have a UTxO that simultaneously has _both_ a lot of dust _and_ a lot of useful outputs.

Though perhaps it would make sense to change the order of these two properties around?

We could write:

> A coin selection algorithm should:
> * employ strategies to limit the amount of _dust_ that accumulates in the UTxO set.
> * over the course of time, aim to generate and maintain a UTxO set with _useful_ outputs: that is, outputs that allow us to process future payments with a _reasonably small_ number of inputs.

Thoughts?

information/specification.md
jonathanknowles: @jonathanknowles wrote:
> 

Perhaps add a short paragraph here explaining why Cardano Wallet gives priority to Random-Improve. This paragraph could link to the "Self Organization in Coin Selection" blog article (listed at the end as an external resource).

Fixed in ebfd135cbbfe2bf7aff749d326f30deca462397f.

information/specification.md
```diff
@@ -0,0 +1,706 @@
...
+The [accumulated coin selection](#accumulated-coin-selection-1) is returned
+to the caller as the [coin selection](#coin-selection) result.
+
+The [available UTxO set](#available-utxo-set) is returned to the caller as the
+[remaining UTxO set](#remaining-utxo-set) result.
+
+# External Resources
+
+## Self Organization in Coin Selection
+
```
jonathanknowles: Suggestion: add a short summary of the content of this resource, to explain to readers why they should want to read it.

information/specification.md
jonathanknowles: @rvl wrote:
> What is the difference between fully depleted and insufficient balance?

Good question!

If the balance of a UTxO is insufficient to pay for the requested outputs, then that is a condition that we can detect _immediately_. It's equivalent to `sumValue utxo < sumValue outputs` (assuming a suitable `sumValue` function).

However, even if the balance _appears_ to be sufficient at the outset,  coin selection algorithms can _still_ end up exhausting all of the UTxO entries, due to the fact that coin selection algorithms don't always choose UTxO entries in a way that is "optimal". One complication here is the self-imposed restriction that any given UTxO entry can only be used to pay for a single output.

@KtorZ If we do run into the `UTxOFullyDepleted` condition (which I suspect is rare), is there a good reason to not relax the above restriction, so that, if needed, we allow a UTxO entry to pay for more than one output? This could be a very simple algorithm that doesn't attempt to perform any other optimization.

information/specification.md
jonathanknowles: > > Since a single UTxO entry can only be used to pay for a single output

@KtorZ wrote:
> This is **not** generally true and in particular, not true for Cardano. This is a limitation our coin selection algorithm self-imposed to itself to make the problem easier to solve (so a multi-variable problem becomes multiple independant single-variable problems, easier to solve).

**Hypothetical question:** is there a good reason that we could not introduce (in some future version of the wallet) a **_third algorithm_** (to fall back on after trying `Random-Improve` and `Largest-First`) that _relaxes_ the above restriction, and just attempts to fulfill the payment (without attempting any other optimizations)?

information/specification.md
jonathanknowles: @rvl I've made an additional change to clarify the difference here: 327efe4e07672933777a6b775463148e3817b1bc

Let me know if you think it can be clarified further!

information/specification.md
jonathanknowles: >> Since a single UTxO entry can only be used to pay for a single output

> This is **not** generally true and in particular, not true for Cardano. 

I've made a change which fixes this here: 02711d3cd95a6d5cb446b7239732e4a95d80e2ee.

information/specification.md
jonathanknowles: @KtorZ wrote:
> I wouldn't put too much emphasis on wallets in this document. 

I agree that we should avoid making the descriptions of these algorithms depend on the details of wallets. Though perhaps there is a balance that we can strike here?

Of course, the concept of a coin selection algorithm can be completely divorced from the concept of a wallet. So we take a lot of care, for example, to talk about the `initial UTxO set` rather than the `wallet UTxO set`.

But perhaps it helps to occasionally mention the concept of a wallet, as that is a very common context in which a UTxO appears, and it gives readers a bit of context for how these algorithms are used.

How about the following **alternative wording**:

> In the context of a _wallet_, if a coin selection algorithm is applied to the wallet's _complete_ UTxO set, then the remaining UTxO set represents the _updated_ UTxO set of that wallet.

information/specification.md
KtorZ: > Hypothetical question: is there a good reason that we could not introduce (in some future version of the wallet) a third algorithm (to fall back on after trying Random-Improve and Largest-First) that relaxes the above restriction, and just attempts to fulfill the payment (without attempting any other optimizations)?

In theory, no, but there has been a fair amount of research and simulations done regarding the current algorithms. Thus introducing a third one isn't necessarily a piece of cake; now, we'd need to see the influence of a third component on the UTxO

information/specification.md
jonathanknowles: @jonathanknowles wrote:
> Suggestion: add a short summary of the content of this resource, to explain to readers why they should want to read it.

Fixed in b778151bdb1eed088fbf83be66b210f43c9b066c.

information/specification.md
jonathanknowles: @rvl wrote:
> Any reason why this is hash and not transaction id, which is more general?

@KtorZ also wrote:
> it doesn't quite matter how h is calculated here (a hash, an encrypted digest .. whatever) but what's important though is that h must be a unique transaction identifier.

Thanks for this. I've fixed this in the following change: a76cc0b6201e3e52cef6c2239517558621d7502d

I've also added this to the [list of types that can be generalized](https://github.com/input-output-hk/cardano-coin-selection/issues/17) within `cardano-coin-selection`.

information/specification.md
jonathanknowles: > Shouldn't the document actually starts with the definitions? 

That's a good question.

My original intention was to have a brief, human-readable introduction that _doesn't_ rely on the reader to first have to read through a set of definitions.

The question I'm now considering is this: would a reader that is representative of our target audience be able to comprehend the introduction section _without_ too much difficulty, and _without_ having to jump to the definitions section?

One possible alternative solution: we could add more **hyperlinks** from terms to their definitions, as we have done in other places. Thoughts?

information/specification.md
```diff
@@ -0,0 +1,733 @@
...
+# Background
+
+## What is Coin Selection?
+
+Coin selection is the process of choosing unspent coins from a wallet in order
+to pay money to one or more recipients.
+
+In the world of physical money, wallets hold value in the form of _coins and
+banknotes_.  When making a payment, we typically select a number of coins and
```
Anviking: The "typically" here makes me think of: "what is the alternative to selecting a number of coins and banknotes", and "is it true that I / one typically does this?", which I think distracts me from the point.

I wonder if it wouldn't be more effective to use a specific concrete example instead of having to _correctly_ generalise the typical scenario.

E.g, rather than
```
When making a payment [...] 
we typically select [...] 
it's often impossible to select the exact amount [...] "
in such cases we typically give the recipient more
```
what about something more like
```
Alice buys food [...] 
she selects [...] 
she cannot select the exact amount [...]
she pays more than"
```
?

information/specification.md
jonathanknowles: @KtorZ wrote:
> A UTxO is actually a transaction identifier and an output index. I am not sure to agree with the definition here

Fair enough!

When I originally wrote this article, I was attempting to be consistent with the types defined in our code, which _effectively_ defines the UTxO set as a set of pairs of (`TxIn`, `TxOut`):

```hs
newtype UTxO = UTxO { getUTxO :: Map TxIn TxOut }
```

However, as far as coin selection algorithms are concerned, the `UTxO` could also be defined as:

```hs
newtype UTxO k = UTxO { getUTxO :: Map k Coin }
```

From the point of view of a coin selection algorithm, so long as `k` is instantiated to a type that provides an `Ord` instance, the other details are irrelevant. (See https://github.com/input-output-hk/cardano-coin-selection/pull/29.)

The question then becomes:
> How can we relate this type to the human-readable definition that we provide?

We could write something like this:

> A _UTxO set_, or the set of _unspent transaction outputs_, represents the unspent value associated with a wallet.
>
> From the point of view of a coin selection algorithm, each entry in the UTxO set consists of a pair of the form (**_o_**, **_v_**), where:
>   * **_o_** is a [transaction output](#transaction-output);
>   * **_v_** is a [coin value](#coin-value).

Which can then reuse our _existing_ definition of _transaction output_:
> a pair (**_h_**, **_n_**), where:
>  * **_h_** is a _unique identifier_ for an existing transaction _t_;
 > * **_n_** is a 0-based integer index into the output list of transaction _t_.

Thoughts?

information/specification.md
jonathanknowles: > I wonder if it wouldn't be more effective to use a specific concrete example instead of having to correctly generalise the typical scenario.

Perhaps we could provide both? Some people do seem to prefer general definitions, but it's clear that many people also like examples.

information/specification.md
jonathanknowles: See discussion here: https://github.com/input-output-hk/cardano-coin-selection/pull/16#discussion_r400068972

information/specification.md
jonathanknowles: Fixed in https://github.com/input-output-hk/cardano-coin-selection/pull/16/commits/638cd7749ea454037c0f835be180bbd654f73cdd.

information/specification.md
jonathanknowles: I've made the modification indicated above here https://github.com/input-output-hk/cardano-coin-selection/pull/16/commits/31bf5407abf07e6a015f2d5e48f47994dc80a5c3.

information/specification.md
```diff
@@ -0,0 +1,826 @@
...
+is _conserved_ in the [coin selection](#coin-selection) result.
+
+In particular, the _outputs_ field of the [coin selection](#coin-selection)
+result should be _equal to_ the [requested output list](#requested-output-list).
+
+## Failure Modes
+
+There are a number of ways in which a coin selection algorithm can fail:
+
+  * #### UTxO Balance Insufficient
```
piotr-iohk, line 479: Does "Not enough money to cover fee" fall into that category, or maybe it could be mentioned as separate one? (Just wondering... :thinking: )

piotr-iohk (approved):
:100:

information/specification.md
jonathanknowles: @KtorZ I've now completely revised this section with new definitions.

See: https://github.com/input-output-hk/cardano-coin-selection/pull/16/commits/4f266317a99ca3ccf87f1c98a302bb3f617f2a25

information/specification.md
jonathanknowles: @KtorZ I've now updated this section with the alternative wording indicated above.

See here: 874538b1e28327b8004d9eaef83113d159c019e6.

information/specification.md
jonathanknowles: @KtorZ I've now significantly revised both the background and definitions sections.

I've tried to keep the background section self-contained, so that it is readable by someone who hasn't necessarily read the definitions.

Though where appropriate, I've added hyperlinks from the background to the definitions, so the reader can gain greater understanding when they feel it is necessary.

Feel free to let me know if you would like to see further revisions!

information/specification.md
jonathanknowles: Hi @piotr-iohk 
> Does "Not enough money to cover fee" fall into that category, or maybe it could be mentioned as separate one? (Just wondering... thinking )

That's a good point!

I've not yet added the section on fee balancing. But that will include the error you mentioned (as it falls into a separate category of errors).

information/specification.md
jonathanknowles: Closing this section after internal discussion with @rvl, as it seems we really do need to have both of the following:

* **human-language explanations**, to help casual readers gain an intuition of what each property is for.
* **mathematical laws**, to help people that want to re-implement these algorithms.

It may be the case that in future we'll write a more formal specification for coin selection, but that shouldn't necessarily restrict the way we write this document.

information/specification.md
```diff
@@ -0,0 +1,826 @@
...
+    entries in the [requested output list](#requested-output-list), due to
+    various restrictions that coin selection algorithms impose on themselves
+    when selecting UTxO entries.
+
+  * #### Maximum Input Count Exceeded
+
+    This failure occurs if the _number_ of UTxO entries needed to pay for the
+    outputs in the [requested output list](#requested-output-list) exceeds the
+    upper limit specified by the [maximum input count](#maximum-input-count)
+    parameter.
```
KtorZ, line 511: The maximum input count parameter is more of a symptom IMO. This doesn't really tell you why it is wrong to select too many entries (although mentioned in previous sections). What about:

"This failure occurs when another input must be selected by the algorithm to continue but doing so will increase the size of the resulting transaction beyond an acceptable limit."

information/specification.md
```diff
@@ -0,0 +1,826 @@
...
+
+### Conservation of Outputs
+
+This property states that the [requested output list](#requested-output-list)
+is _conserved_ in the [coin selection](#coin-selection) result.
+
+In particular, the _outputs_ field of the [coin selection](#coin-selection)
+result should be _equal to_ the [requested output list](#requested-output-list).
+
+## Failure Modes
```
KtorZ, line 475: It feels odd to have both the possible failures and the properties described before even talking about the algorithms themselves. What about changing the order of the sections here so that we first describe the algorithms, and then, talk about the properties they have. 

It's like going through the troubleshooting section of an installation manual directly without even having looked at the installation instructions in the first place.

information/specification.md
```diff
@@ -0,0 +1,826 @@
...
+The **Largest-First** coin selection algorithm processes
+[outputs](#requested-output-list) in _descending order of coin value_, from
+_largest_ to _smallest_.
+
+For each output, it repeatedly selects the _largest remaining_ unspent UTxO
+entry until the value of selected entries is greater than or equal to the
+value of that output.
+
+The name of the algorithm is taken from the idea that the **largest** UTxO
+entry is always considered **first**.
```
KtorZ, line 546: It'll be good to give a disclaimer here about why is Largest First used as (and only used at) a fallback. There are two main reasons:

a) If we fail to cover a payment when selecting coins from the largest first, then **necessarily**, it is impossible to cover such payment with the given UTxO set.  <-- and this is an actual property of the Largest First algorithm.

b) Always selecting largest coins first to pay for transaction has dramatic effects on the UTxO in the long run for it tends to merge large UTxOs together (reducing payment concurrency) and produces dust (increasing cost of making transactions and possibly making wallets unusable). It is thereby not a viable option to use on a typical use-case.

Also about a), note that our _current implementation_ has an important flaw: it does select largest coins first, but consider each output independently instead of considering the total balance. This is **wrong**. To be honest I thought we fixed that but in doubt, I double-checked and, we actually haven't :no_mouth:

information/specification.md
```diff
@@ -0,0 +1,826 @@
...
+
+### Cardinality
+
+The Largest-First algorithm imposes the following cardinality restriction:
```
KtorZ, line 550: See my previous comment, this restriction should only apply to the Random-Improve algorithm!

information/specification.md
```diff
@@ -0,0 +1,826 @@
...
+
+## Random-Improve
+
+The **Random-Improve** coin selection algorithm works in _two phases_.
+
+In the _first_ phase, the algorithm iterates through each of the [requested
+outputs](#requested-output-list) in _descending order of coin value_, from
+_largest_ to _smallest_. For each output, the algorithm repeatedly selects
+entries at **random** from the [initial UTxO set](#initial-utxo-set), until the
+_total value_ of selected entries is enough to pay for that ouput.
```
KtorZ, line 654: for that _ouput_ → for that _ou**t**put_

information/specification.md
```diff
@@ -0,0 +1,826 @@
...
+
+In the _second_ phase, the algorithm attempts to **improve** upon each of the
+UTxO selections made in the previous phase, by conservatively expanding the
+selection made for each output, in order to generate improved change values.
```
KtorZ, line 658: It'll be worth mentioning here what a "improved" output is and why is that so:

- A new change value is considered an improvement if its value gets _closer_ to the value of the corresponding output being covered in the first place. 

The algorithm works by considering, for each output, the triple (minValue, idealValue, maxValue) of values to cover. `minValue` is obviously the value of the selected output itself and, to be considered successful, the algorithm **must* select inputs to cover for at least this `minValue`. And then, improve to try covering for `idealValue`, but no more than `maxValue` (otherwise, we would end up selecting a lot more than what's really needed). In practice, `idealValue` is set to twice the output value and `maxValue` to thrice the output value. 
This therefore aims for constructing change addresses that are roughly equal to the output value (with an error margin or +/- 1 output value; which turns out to be good enough in simulations and empirical observations). 
On large UTxO sets where the variety of coins is bigger, one could imagine using more restrictive constraints like:

- min = 1.75 * output
- ideal = 2*output 
- max = 2.25 * output

such that change outputs are roughly of the same size as the output but with a greater precision.

information/specification.md
```diff
@@ -0,0 +1,826 @@
...
+
+This phase ends when _every_ output has been associated with a selection of
+UTxO entries.
+
+#### Phase 2: Improvement
+
+In this phase, the algorithm attempts to improve upon each of the UTxO
+selections made in the previous phase, by conservatively expanding the
+selection made for each output in order to generate improved change
+values.
```
KtorZ, line 750: It could be nice to point out that this phase can not make the algorithm fails and is totally optional. As per the 3rd principle, failing to improve _probably_ means that there's no need to improve in the first place which is part of the beauty of this whole algorithm.

information/specification.md
```diff
@@ -0,0 +1,826 @@
...
+
+  * incrementally populates the
+    [accumulated coin selection](#accumulated-coin-selection-1).
+
+For each output of value **_v_**, the algorithm:
+
+ 1.  **Calculates a _target range_** for the total value of inputs used to
+     pay for that output, defined by the triplet:
+
+     (_minimum_, _ideal_, _maximum_) = (_v_, _2v_, _3v_)
```
KtorZ, line 767: Okay so this is mentioned here. I think it'll deserve a small mention in the overview above, at least to explain the goal of the improvement step: getting change values closer to the output value.

information/specification.md
```diff
@@ -0,0 +1,826 @@
...
+
+As mentioned in the [Goals](#goals) section, it is desirable that coin
+selection algorithms, over time, are able to create UTxO sets that have
+_useful_ outputs: outputs that will allow us to process future payments with a
+_reasonably small_ number of inputs.
+
+If for each payment request of value **v** we create a change output of
+_roughly_ the same value **v**, then we will end up with a distribution of
+change values that matches the typical value distribution of payment
+requests.
```
KtorZ, line 689: This one maybe deserves a small example? 

"To give a concrete example: if Alice often buys bread for 1€, then her wallet will self-organize itself over time to contain a lot of 1€ coins."

rvl (approved):
:star: Great doc, this revision is even better than the previous.

information/specification.md
```diff
@@ -0,0 +1,826 @@
...
+    It represents the set of values that remain after the coin selection
+    algorithm has removed values to pay for entries in the [requested output
+    list](#requested-output-list).
+
+    In the context of a _wallet_, if a coin selection algorithm is applied to
+    the wallet's _complete_ UTxO set, then the _remaining_ UTxO set represents
+    the _updated_ UTxO set of that wallet.
+
+## Properties
+
```
jonathanknowles, line 384: We should probably insert a short paragraph here to explain what this section is about.

jonathanknowles: > We should probably insert a short paragraph here to explain what this section is about.

Fixed in e82a7641abf2c09bdb967a7558e89a6ab196111c.

information/specification.md
jonathanknowles: > for that _ouput_ → for that _ou**t**put_

Fixed in e252dd3c33f9a076c2a08a3ed0d74a494c491880.

information/specification.md
jonathanknowles: @KtorZ wrote:
> Also about a), note that our current implementation has an important flaw: it does select largest coins first, but consider each output independently instead of considering the total balance. This is wrong. To be honest I thought we fixed that but in doubt, I double-checked and, we actually haven't :no_mouth: 

I've created a bug report for this here: https://github.com/input-output-hk/cardano-coin-selection/issues/39

information/specification.md
```diff
@@ -0,0 +1,830 @@
...
+Addresses are typically owned (and generated) by individual wallets.
+
+In general, coin selection algorithms are agnostic to the choice of address
+format. Therefore, the details of individual address formats are not described
+within this document.
+
+### Coin Value
+
+A _coin value_ is a positive integer value that represents a number of
```
jonathanknowles, line 211: This should be changed back to **_non-negative_**.

See [this comment](https://github.com/input-output-hk/cardano-coin-selection/issues/38#issuecomment-618262828) on issue https://github.com/input-output-hk/cardano-coin-selection/issues/38.

information/specification.md
jonathanknowles: > It feels odd to have both the possible failures and the properties described before even talking about the algorithms themselves. What about changing the order of the sections here so that we first describe the algorithms, and then, talk about the properties they have.

Hmm. I'm not sure I agree here.

I would argue that the set of potential failure modes is actually part of the API, as it represents a set of errors common to all algorithms. The API itself defines the set of potential errors. (This is also the case in the Haskell library.)

Similarly, the properties described in the properties section apply to all possible algorithms (and not just those we have provided). They specify additional restrictions on the relationships between parameters and results that are not practical to express in the types themselves.

The separation between API and implementation actually mirrors the separation we have in the library itself.

information/specification.md
jonathanknowles: > This failure occurs when another input must be selected by the algorithm to continue but doing so will increase the size of the resulting transaction beyond an acceptable limit.

Thanks. I've incorporated this suggestion into the following change: e31e212078ded1f1b59804f245d4d8c6683b07b5

information/specification.md
jonathanknowles: Fixed in d35f56c06f8924c403b915317e967ea84d212801.

information/specification.md
jonathanknowles: Fixed in d35f56c06f8924c403b915317e967ea84d212801.

information/specification.md
jonathanknowles: @KtorZ I've extended the description of the algorithm in 4cf7b1334a40d382f2642ca1187a9f12a339f8d6.

The above change attempts to explain the motivation behind attempting to produce a selection for each output that's around twice the value of that output.

Let me know if you think it can be improved further. (I'll also make changes to subsequent sections, where appropriate.)

information/specification.md
jonathanknowles: Done! See 4cf7b1334a40d382f2642ca1187a9f12a339f8d6.

information/specification.md
jonathanknowles: > This one maybe deserves a small example?
> 
> "To give a concrete example: if Alice often buys bread for 1€, then her wallet will self-organize itself over time to contain a lot of 1€ coins."

Done! Added in 3471486a2031a36a47d9dd42d12aad897c4027e2.

PR #25, opened 2020-03-24
Extend tests and documentation for fee distribution function.

jonathanknowles:
## Related Issue

In preparation for further work on #21.

## Summary

This PR:

- [x] enhances the Haddock documentation for the fee distribution function, and adds a few worked examples;
- [x] adds a **property test** to verify that the fee distribution is **fair**.
- [x] changes the function type so that only **non-empty lists** are accepted;
- [x] removes the test for the non-empty list precondition (as it is now redundant);
- [x] documents the internals of the function in a way that can be easily translated to the specification document.

### DIFF
--- a/src/Cardano/Fee.hs
+++ b/src/Cardano/Fee.hs
@@ -273,7 +273,7 @@ reduceSingleChange (Fee fee, Coin chng)
 --
 -- == Pre-condition
 --
--- Every coin in the given list must be __non-zero__.
+-- Every coin in the given list must be __non-zero__ in value.
 --
 -- == Examples
 --
@@ -290,11 +290,15 @@ reduceSingleChange (Fee fee, Coin chng)
 -- [(Fee 2, Coin 1), (Fee 4, Coin 2), (Fee 8, Coin 4)]
 --
 distributeFee :: Fee -> NonEmpty Coin -> NonEmpty (Fee, Coin)
-distributeFee _ coins | Coin 0 `F.elem` coins =
-    error "distributeFee: one or more coins has a value of zero."
-distributeFee (Fee feeTotal) coins =
+distributeFee (Fee feeTotal) coinsUnsafe =
     NE.zip feesRounded coins
   where
+    -- A list of coins that are non-zero in value.
+    coins :: NonEmpty Coin
+    coins =
+        invariant "distributeFee: all coins must be non-zero in value."
+        coinsUnsafe (Coin 0 `F.notElem`)
+
     -- A list of rounded fee portions, where each fee portion deviates from the
     -- ideal unrounded portion by as small an amount as possible.
     feesRounded :: NonEmpty Fee
--- a/src/Cardano/Fee.hs
+++ b/src/Cardano/Fee.hs
@@ -290,7 +290,7 @@ reduceSingleChange (Fee fee, Coin chng)
 -- [(Fee 2, Coin 1), (Fee 4, Coin 2), (Fee 8, Coin 4)]
 --
 distributeFee :: Fee -> NonEmpty Coin -> NonEmpty (Fee, Coin)
-distributeFee _ outs | Coin 0 `F.elem` outs =
+distributeFee _ coins | Coin 0 `F.elem` coins =
     error "distributeFee: one or more coins has a value of zero."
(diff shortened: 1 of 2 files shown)


test/Cardano/FeeSpec.hs
```diff
@@ -133,7 +139,7 @@ spec = do
             }) (Right $ FeeOutput
             { csInps = [20,20]
             , csOuts = [17,18]
-            , csChngs = [1,2]
+            , csChngs = [2,1]
```
jonathanknowles: This line fixes a spurious test failure.

The order here is affected by an arbitrary choice of which change entry to round up, and which to round down.

src/Cardano/Fee.hs
```diff
@@ -241,9 +251,12 @@ rebalanceChangeOutputs :: FeeOptions -> Fee -> [Coin] -> [Coin]
 rebalanceChangeOutputs opt totalFee chgs =
     case removeDust (Coin 0) chgs of
         [] -> []
-        xs -> removeDust (dustThreshold opt)
+        x : xs ->
```
jonathanknowles: Here we can take advantage of the fact that we already have a non-empty list.

(We have already excluded the possibility of an empty list in the pattern match above this one.)

src/Cardano/Fee.hs
```diff
@@ -254,29 +267,94 @@ reduceSingleChange (Fee fee, Coin chng)
...
+--
+-- >>> distributeFee (Fee 7) [(Coin 1), (Coin 2), (Coin 4)]
+-- [(Fee 1, Coin 1), (Fee 2, Coin 2), (Fee 4, Coin 4)]
+--
+-- >>> distributeFee (Fee 14) [(Coin 1), (Coin 2), (Coin 4)]
+-- [(Fee 2, Coin 1), (Fee 4, Coin 2), (Fee 8, Coin 4)]
+--
+distributeFee :: Fee -> NonEmpty Coin -> NonEmpty (Fee, Coin)
+distributeFee _ outs | Coin 0 `F.elem` outs =
+    error "distributeFee: one or more coins has a value of zero."
```
paweljakubas: maybe it is worth to use `invariant` here as we did extensively in cardano-wallet core?

src/Cardano/Fee.hs
```diff
@@ -254,29 +267,94 @@ reduceSingleChange (Fee fee, Coin chng)
...
+      where
+        indices :: NonEmpty Int
+        indices = 0 :| [1 ..]
+
+    -- A list of fee portion adjustments, with one adjustment per coin.
+    --
+    -- Since the ideal fee portion for each coin is a rational value, we must
+    -- therefore round the rational value to produce a final integer result.
+    --
+    -- * A fee adjustment of '1' indicates a fee portion to be rounded /up/.
```
paweljakubas: At first look I thought maybe RoundUp, RoundDown would be better, but 1, 0 is also clear here

jonathanknowles: > At first look I thought maybe RoundUp, RoundDown would be better, but 1, 0 is also clear here

I really like this suggestion. :+1: 

I've applied the suggestion here: https://github.com/input-output-hk/cardano-coin-selection/pull/25/commits/a8ce44c9ac76a5eb0132d40887556cb9f34726b5

test/Cardano/FeeSpec.hs
```diff
@@ -430,44 +439,56 @@ propDivvyFee prop (fee, NonEmpty outs) =
             ]
         $ prop (fee, outs)
 
+-- | Verify that fees are distributed fairly across outputs, so that every
+--   rounded fee portion is always within unity of the ideal unrounded fee
+--   portion.
+propDistributeFeeIsFair
```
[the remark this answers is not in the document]
jonathanknowles: Thanks! I think there are a few more properties we could add, but I'll add them in a future PR.

src/Cardano/Fee.hs
jonathanknowles: > maybe it is worth to use `invariant` here as we did extensively in cardano-wallet core?

Fixed in https://github.com/input-output-hk/cardano-coin-selection/pull/25/commits/197433953777c47278a659f44caba6f734df103e.

PR #26, opened 2020-03-25
Add property to verify that fee distributions are optimal.

jonathanknowles:
## Related Issue

Preparation for #21.

## Summary

This PR adds a simple property to verify that the fee distribution produced by `distributeFee` is optimal.

Since `distributeFee` is restricted to generating **integer** fee distributions, for any given distribution there will be some deviation from the ideal **unrounded** fee distribution, caused by the particular choice of rounding.
    
This property verifies that the absolute deviation from the ideal unrounded fee distribution is minimal across all possible ways to round.

### DIFF
--- a/test/Cardano/FeeSpec.hs
+++ b/test/Cardano/FeeSpec.hs
@@ -346,7 +346,7 @@ spec = do
 
     describe "distributeFee" $ do
         it "fee portions are all within unity of ideal unrounded portions"
-            (checkCoverage propDistributeFeeIsFair)
+            (checkCoverage propDistributeFeeFair)
         it "Σ fst (distributeFee fee outs) == fee"
             (checkCoverage propDistributeFeeSame)
         it "snd (distributeFee fee outs) == outs"
@@ -442,10 +442,10 @@ propDistributeFee prop (fee, outs) =
 -- | Verify that fees are distributed fairly across outputs, so that every
 --   rounded fee portion is always within unity of the ideal unrounded fee
 --   portion.
-propDistributeFeeIsFair
+propDistributeFeeFair
     :: (Fee, NonEmpty Coin)
     -> Property
-propDistributeFeeIsFair (fee, coins) = (.&&.)
+propDistributeFeeFair (fee, coins) = (.&&.)
     (F.all (uncurry (<=)) (NE.zip fees feeUpperBounds))
     (F.all (uncurry (>=)) (NE.zip fees feeLowerBounds))
   where
--- a/test/Cardano/FeeSpec.hs
+++ b/test/Cardano/FeeSpec.hs
@@ -36,6 +36,8 @@ import Cardano.Types
     )
 import Control.Arrow
     ( left )
+import Control.Monad
+    ( replicateM )
 import Control.Monad.IO.Class
     ( liftIO )
 import Control.Monad.Trans.Except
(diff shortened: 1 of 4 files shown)


test/Cardano/FeeSpec.hs
```diff
@@ -460,6 +466,54 @@ propDistributeFeeIsFair (fee, coins) = (.&&.)
...
+      where
+        -- Indicates whether the given distribution has the correct total fee.
+        isValidDistribution :: NonEmpty Fee -> Bool
+        isValidDistribution r = sum (getFee <$> r) == getFee fee
+
+        -- All possible ways to round an unrounded fee distribution.
+        allPossibleRoundings :: [NonEmpty RoundingDirection]
+        allPossibleRoundings = traverse (const [RoundUp, RoundDown]) coins
+
+    -- The ideal unrounded fee distribution.
```
paweljakubas: so ideal fee is when we treat (ie., sum) all coins as one coin and calculate fee on it. not afraid here we will exceed 45b ada and it will affect somehow this?

test/Cardano/FeeSpec.hs
```diff
@@ -677,7 +731,7 @@ instance Arbitrary FeeOptions where
             }
 
 instance Arbitrary a => Arbitrary (NonEmpty a) where
-    arbitrary = (:|) <$> arbitrary <*> arbitrary
+    arbitrary = (:|) <$> arbitrary <*> (take 10 <$> arbitrary)
```
paweljakubas: why not use : 
```
num <- choose (1,100) 
```
and use instead 10?

test/Cardano/FeeSpec.hs
```diff
@@ -460,6 +466,54 @@ propDistributeFeeIsFair (fee, coins) = (.&&.)
...
+        (\f roundDir -> Fee $ fromIntegral @Integer $ round roundDir f)
+        idealUnroundedDistribution <$> allPossibleRoundings
+      where
+        -- Indicates whether the given distribution has the correct total fee.
+        isValidDistribution :: NonEmpty Fee -> Bool
+        isValidDistribution r = sum (getFee <$> r) == getFee fee
+
+        -- All possible ways to round an unrounded fee distribution.
+        allPossibleRoundings :: [NonEmpty RoundingDirection]
+        allPossibleRoundings = traverse (const [RoundUp, RoundDown]) coins
```
paweljakubas: :+1:

test/Cardano/FeeSpec.hs
```diff
@@ -460,6 +466,54 @@ propDistributeFeeIsFair (fee, coins) = (.&&.)
...
+    -- The minimum deviation across all possible distributions for the given
+    -- fee and set of coins.
+    minimumPossibleDeviation :: Rational
+    minimumPossibleDeviation =
+        F.minimum $ computeDeviation <$> allPossibleDistributions
+
+    -- The set of all possible fee distributions for the given fee and coins.
+    allPossibleDistributions :: [NonEmpty Fee]
+    allPossibleDistributions = filter isValidDistribution $ NE.zipWith
+        (\f roundDir -> Fee $ fromIntegral @Integer $ round roundDir f)
```
paweljakubas: ok, so the idea is to have ideal distribution, and then *perturb* it using all possible roundings, and then filter out valid distributions.

jonathanknowles: > ok, so the idea is to have ideal distribution, and then **apply perturbation** it using all possible roundings, and then filter out valid distributions.

Yes, exactly!

test/Cardano/FeeSpec.hs
jonathanknowles: > not afraid here we will exceed 45b ada and it will affect somehow this?

I think this shouldn't be a problem, as:

1. all internal calculations use either `Integer` or `Rational`, only converting back to `Fee` at the end.
2. all the `Fee` portions should be less than or equal to the initial `Fee` value.

test/Cardano/FeeSpec.hs
jonathanknowles: Fixed in https://github.com/input-output-hk/cardano-coin-selection/pull/26/commits/efcc8d7019c6a19e3a5d216002ac047208a3c115!

PR #27, opened 2020-03-25
Add property tests for dust coalescing.

jonathanknowles:
## Related Issue

Preparation for #21.

## Summary

This PR:

- [x] **Renames** `removeDust` to `coalesceDust`.
    **Justification**: this function _doesn't_ always succeed at removing all dust, but it _does_ always succeed at coalescing dust to the greatest extent possible.
- [x] **Adds property tests** to check that:
    - [x] `coalesceDust` preserves the _total sum_.
    - [x] `coalesceDust` removes _all_ zero-valued coins.
    - [x] `coalesceDust` leaves behind _at most_ one dust coin.
- [x] **Corrects errors** in the documentation for `coalesceDust` and `dustThreshold`, emphasizing that the dust threshold value defines an _inclusive_ upper limit for dust values, rather than an _exclusive_ upper limit.

### DIFF
--- a/test/Cardano/FeeSpec.hs
+++ b/test/Cardano/FeeSpec.hs
@@ -55,6 +55,8 @@ import Data.Functor.Identity
     ( Identity (runIdentity) )
 import Data.List.NonEmpty
     ( NonEmpty (..) )
+import Data.Monoid
+    ( All (..) )
 import Data.Ratio
     ( (%) )
 import Data.Word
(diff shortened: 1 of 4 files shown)


test/Cardano/FeeSpec.hs
```diff
@@ -546,6 +563,72 @@ propDistributeFeeNoNullFee (fee, outs) =
...
+
+propCoalesceDustLeavesNoZeroCoins :: CoalesceDustData -> Property
+propCoalesceDustLeavesNoZeroCoins (CoalesceDustData threshold coins) =
+    property $
+    cover 4 (F.all  (== Coin 0) coins) "∀ coin ∈ coins . coin = 0" $
+    cover 4 (F.elem    (Coin 0) coins) "∃ coin ∈ coins . coin = 0" $
+    cover 8 (F.notElem (Coin 0) coins) "∀ coin ∈ coins . coin > 0" $
+    F.notElem (Coin 0) $ coalesceDust threshold coins
+
+propCoalesceDustLeavesAtMostOneDustCoin :: CoalesceDustData -> Property
```
paweljakubas: how about checking situation when we all coins are below threshold?

jonathanknowles: Fixed: https://github.com/input-output-hk/cardano-coin-selection/pull/27/commits/b7bf2b6cb60c23f2c17b153c112ed99367fcb511

Issue #22, opened 2020-03-13
Address areas of missing code coverage for coin selection.


jonathanknowles:
# Context

There are a few areas of missing code coverage in the [`cardano-coin-selection`](https://github.com/input-output-hk/cardano-coin-selection) library.

## Areas discovered from mutation testing

- [x] Fix issues outlined in https://github.com/input-output-hk/cardano-coin-selection/issues/8

## Other (potential) areas of missing coverage

- [x] Add a test for the invariant of function `adjustForFee`.
- [x] Add a test for the invariant of function `distributeFee`.
- [x] Add property tests for function `Cardano.Fee.splitCoin`.
- [x] Add property tests for function `Cardano.Fee.reduceChangeOutputs`.
- [x] Add test of fairness for function `Cardano.Fee.distributeFee`.

# Decision

Address these areas of missing code coverage with property test or unit tests, as appropriate.

Issue #18, opened 2020-02-26
Revise documentation for `cardano-coin-selection` library.


jonathanknowles:
# Context

As part of the work to create a separate [**coin selection and fee balancing library**](https://github.com/input-output-hk/cardano-wallet/milestone/58), we'd like the library to be documented in a way that's consistent with good practices used by other popular libraries in the Haskell ecosystem.

# Decision

Revise the public documentation for the [`cardano-coin-selection`](https://github.com/input-output-hk/cardano-coin-selection) library.

# Acceptance Criteria

## Source Code Documentation (Haddock)

- [x] Every public module _should_ have a clear summary to state its purpose.
- [x] Every exported function _should_ have a clear comment to illustrate its purpose. 
- [x] Every exported type _should_ have a clear comment to illustrate its purpose.
- [x] Every exported constructor _should_ have a clear comment to illustrate its purpose.

## Other Documentation

- [x] The [home page](https://github.com/input-output-hk/cardano-coin-selection) for the `cardano-coin-selection` library _must_ explain the basic purpose of the library. This page can delegate to source code documentation where appropriate, to avoid duplication.

# Development 

- [x] Remove Cardano-specific blockchain limitations from comments. For example:
    https://github.com/input-output-hk/cardano-coin-selection/blob/f62d069062f196d3b251ac6e7a039a0030d929d6/src/library/Cardano/CoinSelection/Migration.hs#L19-L21
- [x] State what is meant by `relatively-balanced` in function `idealBatchSize`. See [this comment](https://github.com/input-output-hk/cardano-coin-selection/pull/55#discussion_r413504278). (We eventually solved this by referring to `idealBatchSize` as an automatic way to generate a suitable batch size for `selectCoins`. See https://input-output-hk.github.io/cardano-coin-selection/haddock/cardano-coin-selection-1.0.0/Cardano-CoinSelection-Algorithm-Migration.html#v:selectCoins.)
- [x] Check for obvious grammar and spelling errors.

# QA 

_Section to be added._

piotr-iohk:
@jonathanknowles  haddock doc (https://input-output-hk.github.io/cardano-coin-selection/haddock/) returns 404 for: 

Cardano.CoinSelection.Algorithm
Cardano.CoinSelection.Algorithm.LargestFirst
Cardano.CoinSelection.Algorithm.Migration
Cardano.CoinSelection.Algorithm.Random
Cardano.CoinSelection.Fee

[...]

PR #32, opened 2020-04-06
Clarify documentation and properties of `reduceChangeOutputs`.

jonathanknowles:
## Related Issues

#18
#21
#22 

## Summary

This PR:

- [x] Clarifies the documentation of `reduceChangeOutputs`, adding examples.
- [x] Tests that `reduceChangeOutputs` only produces valid coins.

### DIFF
--- a/src/Cardano/Fee.hs
+++ b/src/Cardano/Fee.hs
@@ -242,19 +242,11 @@ coverRemainingFee (Fee fee) = go [] where
                 Nothing -> do
                     lift $ throwE $ ErrCannotCoverFee (fee - sumInputs acc)
 
--- | Pay for the given fee by subtracting it from the given list of change
+-- | Pays for the given fee by subtracting it from the given list of change
 --   outputs, so that each change output is reduced by a portion of the fee
 --   that's in proportion to its relative size.
 --
--- Any dust outputs in the resulting list are coalesced according to the given
--- dust threshold. (See 'coalesceDust'.)
---
--- If there's not enough change to pay for the fee, or if there's only just
--- enough to cover it, this function returns the /empty list/.
---
--- == Examples
---
--- ==== Proportionality
+-- == Basic Examples
 --
 -- >>> reduceChangeOutputs (DustThreshold 0) (Fee 4) (Coin <$> [2, 2, 2, 2])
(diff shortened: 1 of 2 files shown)


src/Cardano/Fee.hs
```diff
@@ -270,29 +270,51 @@ coverRemainingFee (Fee fee) = go [] where
...
+-- >>> reduceChangeOutputs (DustThreshold 0) (Fee 4) (Coin <$> [2, 2, 2, 2])
+-- [Coin 1, Coin 1, Coin 1, Coin 1]
+--
+-- >>> reduceChangeOutputs (DustThreshold 0) (Fee 15) (Coin <$> [2, 4, 8, 16])
+-- [Coin 1, Coin 2, Coin 4, Coin 8]
+--
+-- ==== Dust Coalescing
+--
+-- >>> reduceChangeOutputs (DustThreshold 1) (Fee 4) (Coin <$> [2, 2, 2, 2])
+-- [Coin 4]
```
jonathanknowles, line 296: Perhaps move this example to the explanation regarding coalescing of dust outputs above.

test/Cardano/FeeSpec.hs
```diff
@@ -679,12 +681,12 @@ instance Arbitrary RebalanceChangeOutputsData where
             , pure coinSum
             , choose (coinSum + 1, coinSum * 2)
             ]
-        pure $ RebalanceChangeOutputsData fee threshold coins
+        pure $ ReduceChangeOutputsData fee threshold coins
     shrink = genericShrink
```
jonathanknowles, line 685: Using `genericShrink` here will eventually produce 0-valued coins, which may or may not be desirable. We should check this.

src/Cardano/Fee.hs
jonathanknowles: Fixed in 32ea20368468602fec0865f17571fd2443a231e0.

test/Cardano/FeeSpec.hs
jonathanknowles: Tested in 5533bf919b5b15dbc719bb3444b223fd4cbbc7f8.

PR #33, opened 2020-04-06
Remove dead code unrelated to coin selection.

jonathanknowles:
## Related Issues

Discovered as a consequence of working on #21.

## Summary

This PR removes the following functions and types:

- [x] `computeFee`
- [x] `FeePolicy`
- [x] `Quantity`

These functions and types are not used by the coin selection or fee balancing code.


src/Cardano/Fee.hs
```diff
@@ -116,27 +106,6 @@ newtype DustThreshold = DustThreshold
...
---
--- where @a@ & @b@ are values fixed by the protocol.
-computeFee
-    :: FeePolicy
-    -> Quantity "byte" Int
-    -> Fee
-computeFee policy (Quantity sz) =
-    Fee $ ceiling (a + b*fromIntegral sz)
-  where
-    LinearFee (Quantity a) (Quantity b) (Quantity _c) = policy
```
jonathanknowles, line 138: I was surprised to find that this function isn't actually used by the fee balancing code.

src/Cardano/Types.hs
```diff
@@ -67,23 +62,6 @@ import Quiet
...
--- | A linear equation of a free variable `x`. Represents the @\x -> a + b*x@
--- function where @x@ can be the transaction size in bytes or, a number of
--- inputs + outputs.
---
--- @a@, @b@ and @c@ are constant coefficients.
-data FeePolicy = LinearFee
-    (Quantity "lovelace" Double)
-    (Quantity "lovelace/byte" Double)
-    (Quantity "lovelace/certificate" Double)
-    deriving (Eq, Show, Generic)
```
jonathanknowles, line 83: Given that `computeFee` is not used, this type is therefore also unnecessary.

src/Data/Quantity.hs
```diff
@@ -1,70 +0,0 @@
...
--- >>> b :: Quantity "lovelace/byte" Word32
---
--- so mixing them up is more difficult.
---
--- >>> Aeson.encode $ Quantity @"lovelace" 14
--- {"unit":"lovelace","quantity":14}
-newtype Quantity (unit :: Symbol) a = Quantity { getQuantity :: a }
-    deriving stock (Eq, Generic, Ord)
-    deriving newtype (Bounded, Enum)
-    deriving Show via (Quiet (Quantity unit a))
```
jonathanknowles, line 59: Because `FeePolicy` was the last remaining type to make use of `Quantity`, this type can also be deleted.

src/Cardano/Fee.hs
paweljakubas: yeah, this is very wallet specific. we ingest fee policy and determine `Fee` . And `cardano-coin-selection` starts from this level ;-)

PR #35, opened 2020-04-08
Make `remainingFee` more amenable to property testing.

jonathanknowles:
## Related Issues

#21 
#22 

## Summary

This PR:

- [x] Adds a new type `FeeEstimator`, which wraps the `estimateFee` function. Values of this type can be passed to functions that don't care about `DustThreshold` (instead of `FeeOptions`).
- [x] Adjusts the type of `remainingFee` to accept `FeeEstimator` rather than `FeeOptions`, making it easier to test. (We no longer need to provide a dust threshold, which will be ignored anyway.)
- [x] Adds the type `FeeParameters` for testing, with a function to create a `FeeEstimator` from a `FeeParameters`. Unlike `FeeEstimator`, `FeeParameters` can have a `Show` instance, which makes it much easier to debug failing properties.

- [x] Some related simplifications of the code.

A further PR will provide property tests for `remainingFee`.

### DIFF
--- a/src/library/Cardano/Fee.hs
+++ b/src/library/Cardano/Fee.hs
@@ -420,25 +420,24 @@ remainingFee
     => FeeEstimator i o
     -> CoinSelection i o
     -> Fee
-remainingFee FeeEstimator {estimateFee} s = do
-    if fee >= diff
-    then Fee (fee - diff)
-    else do
+remainingFee FeeEstimator {estimateFee} s
+    | fee >= diff =
+        Fee (fee - diff)
+    | feeDangling >= diff =
+        Fee (feeDangling - fee)
+    | otherwise =
         -- NOTE
         -- The only case where we may end up with an unbalanced transaction is
         -- when we have a dangling change output (i.e. adding it costs too much
         -- and we can't afford it, but not having it result in too many coins
         -- left for fees).
-
-        if (feeDangling >= diff)
-            then Fee (feeDangling - fee)
-            else error $ unwords
-                [ "Generated an unbalanced tx! Too much left for fees"
-                , ": fee (raw) =", show fee
-                , ": fee (dangling) =", show feeDangling
-                , ", diff =", show diff
-                , "\nselection =", pretty s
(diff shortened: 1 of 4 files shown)


src/library/Cardano/Fee.hs
```diff
@@ -102,7 +99,8 @@ newtype Fee = Fee
 
 -- | Defines the maximum size of a dust coin.
 --
--- See function 'dustThreshold' within 'FeeOptions'.
+-- Change values that are less than or equal to this threshold will be
+-- evicted from created transactions.
```
paweljakubas, line 103: you mean changes dusted will not participate in creating the transaction, right?

src/library/Cardano/Fee.hs
```diff
@@ -417,35 +416,32 @@ coalesceDust (DustThreshold threshold) coins =
...
-        if (feeDangling >= diff)
-            then Fee (feeDangling - fee)
-            else error $ unwords
-                [ "Generated an unbalanced tx! Too much left for fees"
-                , ": fee (raw) =", show fee
-                , ": fee (dangling) =", show feeDangling
-                , ", diff =", show diff
-                , "\nselection =", pretty s
-                ]
+        error $ unwords
```
paweljakubas, line 434: would be good to explicitly test it, ie., that this error is emitted

src/library/Cardano/Fee.hs
jonathanknowles: > you mean changes dusted will not participate in creating the transaction, right?

I agree, this wording is a little awkward. Fixed in: 373a41ac334b5e88e51f6fcbd1b60ed66f31432b

src/library/Cardano/Fee.hs
jonathanknowles: > would be good to explicitly test it, ie., that this error is emitted

Definitely! That is actually the subject of PR https://github.com/input-output-hk/cardano-coin-selection/pull/36 (still WIP).

Issue #30, opened 2020-04-02
Revise public API of `cardano-coin-selection` library.


jonathanknowles:
# Context

As part of the work to create a separate [**coin selection and fee balancing library**](https://github.com/input-output-hk/cardano-wallet/milestone/58), we need to ensure that the public API and source code is consistent with best practices generally adopted by the Haskell community.

# Decision

Conduct an examination of the public API and source code, and revise it as necessary to bring it into line with best practices adopted within the Haskell community.

# Development

## Areas to revise

- [x] **Module structure**

    The migration algorithm, while designed to select coins, doesn't have the same type as other algorithms such as `Largest-First` or `Random-Improve`, despite sharing a common `Cardano.CoinSelection` parent module.

    The `Fee` module is also not a coin selection algorithm _per se_, so it should probably not be at the same level as the coin selection algorithms. 

    We decided to adopt the following module hierarchy:

    ```hs
    Cardano
    └── CoinSelection
        ├── Algorithm 
        │   ├── LargestFirst
        │   ├── Migration
        │   └── RandomImprove
        └── Fee
    ```  

- [x] **`Buildable` instances**

    These have been moved out of the public API, and restricted to just the test code.

- [x] **`CoinSelectionError` type**

    We have:
    - [x] Added record fields to disambiguate the constructor arguments.
    - [x] Used specific types for currency values, rather than `Word64`.
    - [x] Embedded more information in each error, where possible.<br><br>

    Fixed in #53.<br><br>

- [x] **`ErrAdjustForFee` type**

    We now:
    - [x] Use the `Fee` type, rather than `Word64`.<br><br>

- [x] **`CoinSelectionAlgorithm` fall-back/delegation**

    The `Random-Improve` algorithm no longer **_delegates_** to `Largest-First` if it fails to find a selection. (See https://github.com/input-output-hk/cardano-coin-selection/pull/44).

    If a library user wishes to implement fallback themselves, it's trivial to implement by calling one algorithm after the other.

    For `cardano-wallet`, it's important that existing fallback behaviour is preserved when we eventually consume the `cardano-coin-selection` from that library. This requirement has been captured on the acceptance criteria here: https://github.com/input-output-hk/cardano-coin-selection/issues/20

- [x] **`Types` module**

    This module has been merged into `CoinSelection`.

- [x] **`ShowFmt` type**

    This has been removed from the public API.

- [x] **`Dom` type**

    This has been removed from the public API.

[...]

PR #37, opened 2020-04-09
Revise Public API: Part 1

jonathanknowles:
## Related Issue

#30 

## Background

This PR is part of a _series_ of PRs aimed at cleaning up the **public API** of `cardano-coin-selection`.

## Summary

This PR makes the following changes:

- [x] Creates an [internal library](https://www.haskell.org/cabal/users-guide/developing-packages.html?highlight=build%20tool%20depends#sublibs) to hold modules that are _not_ intended to be part of the public API. This includes:

    - [x] `Internal.Rounding` (formerly `Numeric.Rounding`)

    - [x] `Internal.Invariant` (contains the `invariant` function)

- [x] Creates test module `Test.Utilities` to hold functions and types _not required_ by our public library but _still required_ by out test suite.

- [x] Moves various functions and types out of `Cardano.Types` and `Cardano.CoinSelection` into `Test.Utilities`. (Note that it _may_ be possible to remove some of these functions and types in future, but this PR focuses on the revising the public API, rather than revising definitions.)

## Additional Minor Changes

- [x] Moves function `distance` from `Types` to `RandomImprove`, as it's only used there.

- [x] Renames `balance` to `utxoBalance`, to distinguish it from other kinds of balance (of which there are several).

- [x] Renames `ErrCoinSelection`to `CoinSelectionError`, so that all types relating to coin selection share a common prefix.


cardano-coin-selection.cabal
```diff
@@ -13,7 +13,7 @@ license-file:        LICENSE
 category:            Web
 build-type:          Simple
 extra-source-files:  README.md
-cabal-version:       >=1.10
+cabal-version:       2.0
```
jonathanknowles, line 16: Moving to a newer `cabal` version is necessary to enable the [internal libraries](https://www.haskell.org/cabal/users-guide/developing-packages.html?highlight=build%20tool%20depends#sublibs) feature.

cardano-coin-selection.cabal
```diff
@@ -40,6 +62,7 @@ library
   build-depends:
       base < 4.14
     , bytestring
+    , cardano-coin-selection-internal
```
jonathanknowles, line 65: The internal library is consumed by the public library.

cardano-coin-selection.cabal
```diff
@@ -74,6 +96,7 @@ test-suite unit
       base
     , bytestring
     , cardano-coin-selection
+    , cardano-coin-selection-internal
```
jonathanknowles, line 99: The internal library is also consumed by our test suite.

cardano-coin-selection.cabal
```diff
@@ -86,12 +109,12 @@ test-suite unit
     , text
     , transformers
     , vector
-  build-tools:
-      hspec-discover
+  build-tool-depends:
+      hspec-discover:hspec-discover
```
jonathanknowles, line 113: Needed to comply with the stricter syntax rules of version 2 of the `cabal` file format.

src/internal/Internal.hs
```diff
@@ -0,0 +1,19 @@
...
+
+-- |
+-- Copyright: © 2018-2020 IOHK
+-- License: Apache-2.0
+--
+-- This module hierarchy provides types and functions that are not intended to
+-- be part of the public API.
+--
+-- Types and functions defined herein are not guaranteed to be forwards or
+-- backwards compatible across different versions of the library.
```
jonathanknowles, line 11: We could actually do without this module, but I thought it useful to include, as it allows us to provide this cautionary message that covers all submodules.

src/library/Cardano/Types.hs
```diff
@@ -0,0 +1,110 @@
...
+--
+newtype Coin = Coin
+    { getCoin :: Word64 }
+    deriving stock (Eq, Generic, Ord)
+    deriving Show via (Quiet Coin)
+
+instance NFData Coin
+
+instance Bounded Coin where
+    minBound = Coin 0
```
jonathanknowles, line 61: I suspect this should be `1`. I've created an issue here: https://github.com/input-output-hk/cardano-coin-selection/issues/38

src/library/Cardano/Types.hs
```diff
@@ -0,0 +1,110 @@
...
+import Quiet
+    ( Quiet (Quiet) )
+
+import qualified Data.Map.Strict as Map
+
+{-------------------------------------------------------------------------------
+                                     Coin
+-------------------------------------------------------------------------------}
+
+-- | A non-negative integer value that represents a number of Lovelace.
```
jonathanknowles, line 49: I suspect this should be *positive*, rather than *non-negative*. I've created an issue here: https://github.com/input-output-hk/cardano-coin-selection/issues/38

cardano-coin-selection.cabal
```diff
@@ -48,17 +71,16 @@ library
     , text
     , transformers
   hs-source-dirs:
-      src
+      src/library
```
paweljakubas, line 74: why not to name `lib` instead of `library` ?

cardano-coin-selection.cabal
```diff
@@ -48,17 +71,16 @@ library
...
       Cardano.CoinSelection
       Cardano.CoinSelection.LargestFirst
       Cardano.CoinSelection.Migration
       Cardano.CoinSelection.RandomImprove
       Cardano.Fee
       Cardano.Types
-      Numeric.Rounding
 
-test-suite unit
+test-suite cardano-coin-selection-test
```
paweljakubas, line 83: when it comes to me I am used to `unit` , `benchmark`, `integration` ... and I like them as they convey what they are in one standarized word

jonathanknowles: >  when it comes to me I am used to unit , benchmark, integration ... and I like them as they convey what they are in one standarized word

It's certainly not fixed in stone.

Though since we already have:
* `cardano-coin-selection`
* `cardano-coin-selection-internal`

It seemed consistent to use:
* `cardano-coin-selection-test`

But we could change this.

src/internal/Internal/Invariant.hs
```diff
@@ -0,0 +1,33 @@
...
+--
+-- If the invariant does not hold (indicated by the predicate function
+-- returning 'False'), throws an error with the specified message.
+--
+-- >>> invariant "not empty" [1,2,3] (not . null)
+-- [1, 2, 3]
+--
+-- >>> invariant "not empty" [] (not . null)
+-- *** Exception: not empty
+invariant
```
paweljakubas, line 24: this is third time I was considering if this should not be some time in the future in `cardano-aux` or `cardano-prelude` if we use it

src/library/Cardano/CoinSelection/RandomImprove.hs
```diff
@@ -395,5 +397,10 @@ mkChange (Output _ (Coin out)) inps =
 -- Utilities
 --------------------------------------------------------------------------------
 
+-- | Compute distance between two numeric values |a - b|
+distance :: (Ord a, Num a) => a -> a -> a
```
paweljakubas, line 401: this could also to the `Internal` or `cardano-aux` ;-)

src/internal/Internal/Invariant.hs
jonathanknowles: > this is third time I am thinking loudly if this should not be some time in the future in cardano-aux or cardano-prelude if we use it

I think you're right. Though I'll leave that one for a future PR. :smile_cat:

src/library/Cardano/CoinSelection/RandomImprove.hs
jonathanknowles: Ah, but that's not exported by `RandomImprove`, so it's fine to leave that where it is.

src/library/Cardano/Types.hs
paweljakubas: for me non-negative is precise. it can be `>= 0` like encoded in `Word64` . no ?

cardano-coin-selection.cabal
jonathanknowles: > why not to name lib instead of library ?

No good reason in particular.

Since everything else (`internal`, `test`, etc.) at this level was not an abbreviation, it seemed fine to use whole words at this level.

But we could change it to `lib` if people feel strongly about it. :+1:

src/test/Test/Utilities.hs
```diff
@@ -6,50 +6,69 @@
...
 {-# LANGUAGE TypeFamilies #-}
 
--- | Utility types used purely for testing.
+-- | Utility functions and types used purely for testing.
 --
 -- Copyright: © 2018-2020 IOHK
 -- License: Apache-2.0
 --
-module Test.Cardano.Types
+module Test.Utilities
```
paweljakubas, line 17: it is quite Cardano specific, I would convey `Cardano` in namescope or at least underline it in comments

src/library/Cardano/Types.hs
jonathanknowles: I mentioned it here as the specification defines it as a positive integer:

https://github.com/input-output-hk/cardano-coin-selection/pull/16/files#diff-f8a2c3e44e2d8bf1d848c3c112859487R211

We already impose an upper bound, which corresponds to 45 million billion Lovelace (45 followed by 15 zeros), which is significantly less than `maxBound :: Word64`.

src/test/Test/Utilities.hs
jonathanknowles: Fixed here: 7fe4b2394c7f9c98f38b3f1ae4e0eb3f273f3475

jonathanknowles:
> My feeling is that maybe all stuff in `Internal` directory should go to just `Internal` module

I can see where you're coming from here, as having one module seems like a simpler solution.

However, I anticipate adding further modules to `Internal` later on (in order to separate out the public and internal parts of the coin selection algorithms) so I think it's still better to separate things out by topic, as we have done at the moment (_rounding numeric values_ and _function invariants_ are quite different topics, for example).

> (and here also add `distance`) instead of being in separate modules

In the case of `distance`, that particular function is just used by `RandomImprove` (and not exported), so I think it makes sense to include it there, rather than defining it somewhere else.

PR #40, opened 2020-04-10
WIP: Selectively export types and functions from public API

jonathanknowles:
## Related Issues

#30

## Summary

Several modules within the coin selection library export functions and types that are:

1. tested by our test suite, **but**
2. not needed in the public API.

This creates a difficulty: we **must** export these functions (from somewhere) in order for the test suite to be able to test them, but we still want to control what we export in the public API.

This PR solves the above problem by:

- [x] Moving main library modules into the existing `cardano-coin-selection-internal` [internal library](https://www.haskell.org/cabal/users-guide/developing-packages.html#sublibs).
- [x] Re-exporting just the parts we need from the main public library.

This solution works well with Haddock, as hyperlinks will only target publicly exported identifiers. (So no accidental linking back to the non-exported parts.)


paweljakubas:
@jonathanknowles I buy into idea that everything in internal can flow, change... and public api should be as stable as possible without breaking changes.  and `Internal` does signaling that I understand. on the other hand we have voices like that https://nikita-volkov.github.io/internal-convention-is-a-mistake/ 
Would be goo to have one consistent "standard" on that and employ throughout all projects (at least new in adrestia realm)
  
Maybe @KtorZ has strong opinion on that. Personally I would stick to what it is and tackle other issues/challenges ;-)

KtorZ:
Internal modules can be extremely annoying when not exposed. Sometimes, you do want to fiddle with the Internal stuff without having to fork a library. 

My own preference is to keep the internal stuff "hidden" from the public API / documentation (see https://github.com/input-output-hk/cardano-transactions/blob/master/src/Data/UTxO/Transaction/Cardano/Byron.hs#L8). Here, some internal functions are still exposed so they can be used if necessary by some privileged piece of code (like the command-line, which definitely knows what to do with these), but they aren't visible on the haddock documentation. 

This is IMO  a reasonable trade-off. Those working from the source are aware that they are in the internal lands. But what's presented through the public interface should indeed be as stable as possible and clear of any unnecessary details.

PR #41, opened 2020-04-13
Revise Public API Types

jonathanknowles:
## Related Issue

#30 

## Summary

This PR:

- [x] Provides a new type `CoinMap`:

    ```hs
    newtype CoinMap a = CoinMap { unCoinMap :: Map a Coin }
    ```

- [x] Uses `CoinMap` to unify and replace the following types:
  - `UTxO`
  - `[Input]`
  - `[Output]`<br><br>

- [x] Uses `CoinMap` to adjust the definition of `CoinSelection`:

    ```patch
      data CoinSelection i o = CoinSelection
    -    { inputs  :: [Input  i]
    -    , outputs :: [Output o]
    +    { inputs  :: CoinMap i
    +    , outputs :: CoinMap o
         , change  :: [Coin]
         }
    ```
    Benefits:

    - [x] the type of `CoinSelection` now prevents duplicate inputs.
    - [x] the type of `CoinSelection` now prevents duplicate outputs.
    - [x] we now no longer need tests to ensure that algorithms are agnostic to ordering.
    - [x] we now no longer need separate types for `Input` and `Output`.<br><br>

- [x] Uses `CoinMap` to adjust the definition of `CoinSelectionAlgorithm`:

    ```patch
    - newtype CoinSelectionAlgorithm i o u m e = CoinSelectionAlgorithm
    + newtype CoinSelectionAlgorithm i o   m e = CoinSelectionAlgorithm
          { selectCoins
              :: CoinSelectionOptions i o e
    -         -> UTxO u
    -         -> NonEmpty (Output o)
    -         -> ExceptT (CoinSelectionError e) m (CoinSelection i o, UTxO u)
    +         -> CoinMap i
    +         -> CoinMap o
    +         -> ExceptT (CoinSelectionError e) m (CoinSelection i o, CoinMap i)
          }
    ```
    Benefits:

    - [x] One fewer type parameter :`u` has been removed (it was always equivalent to `i`).
    - [x] The symmetry between inputs and outputs is clearer: they are both maps.<br><br>

## Remaining Work

- [x] Make the test suite pass.


src/library/Cardano/CoinSelection.hs
```diff
@@ -55,93 +164,62 @@ import GHC.Generics
...
+    }
+    deriving (Generic, Show, Eq)
 
--- NOTE:
---
--- We don't check for duplicates when combining selections because we assume
--- they are constructed from independent elements.
---
--- As an alternative to the current implementation, we could 'nub' the list or
--- use a 'Set'.
```
jonathanknowles, line 89: This assumption is no longer necessary.

However, we should test that:
- [ ] combining selections preserves all keys.
- [ ] combining selections preserves total coin value.

src/library/Cardano/CoinSelection.hs
```diff
@@ -55,93 +164,62 @@ import GHC.Generics
...
---
-instance Semigroup (CoinSelection i o) where
+instance (Ord i, Ord o) => Semigroup (CoinSelection i o) where
     a <> b = CoinSelection
         { inputs = inputs a <> inputs b
         , outputs = outputs a <> outputs b
         , change = change a <> change b
         }
```
jonathanknowles, line 198: There _should_ be an easy way to derive this.

src/library/Cardano/CoinSelection.hs
```diff
@@ -11,38 +13,149 @@
...
+-- | A non-negative integer value that represents a number of Lovelace.
+--
+-- One Ada is equal to 1,000,000 Lovelace.
+--
+newtype Coin = Coin
+    { unCoin :: Word64 }
+    deriving stock (Eq, Generic, Ord)
+    deriving Show via (Quiet Coin)
+
+instance Monoid Coin where
```
paweljakubas, line 80: just nitpicking, but I would put `Semigroup` before `Monoid` ;-)

src/library/Cardano/CoinSelection.hs
```diff
@@ -11,38 +13,149 @@
...
+instance Semigroup Coin where
+    Coin a <> Coin b = Coin (a + b)
+
+instance NFData Coin
+
+instance Bounded Coin where
+    minBound =
+        Coin 0
+    maxBound =
+        Coin 45_000_000_000_000_000
```
paweljakubas, line 92: I always was wondering if this would not be good to have wrapper type on top of `Coin` that encodes ADA/LOVELACE information. Something like Quantity "ada"/Quantity "lovelace" that enables to talk with whatever "level" and always be correct....

src/library/Cardano/CoinSelection.hs
```diff
@@ -11,38 +13,149 @@
...
+    deriving (Eq, Generic)
+    deriving Show via (Quiet (CoinMap a))
+
+instance Foldable CoinMap where
+    foldMap f = F.fold . fmap (f . entryKey) . coinMapToList
+
+instance Ord a => Monoid (CoinMap a) where
+    mempty = CoinMap mempty
+
+instance Ord a => Semigroup (CoinMap a) where
```
paweljakubas, line 115: same nitpicking as above ;-)

src/library/Cardano/CoinSelection.hs
paweljakubas: yeah, probably deft usage of `derivingVia` would do the trick. But on the another hand the current version is so straightforward and easy to grasp

src/library/Cardano/CoinSelection.hs
jonathanknowles: Fixed by the following commits:

45f7021cea5ddaa76ead1c0f3777d3e5413fc11f
2f695cb86b966fafab573644b7f179b37d0c3184

src/library/Cardano/CoinSelection.hs
jonathanknowles: > There _should_ be an easy way to derive this.

If there is an easier way, we can add it to a future PR. :wink:

src/library/Cardano/CoinSelection.hs
jonathanknowles: > just nitpicking, but I would put `Semigroup` before `Monoid` ;-)

I was writing them in alphabetical order! But mathematically, you are absolutely right. :+1:

src/library/Cardano/CoinSelection.hs
jonathanknowles: > I always was wondering if this would not be good to have wrapper type on top of `Coin` that encodes ADA/LOVELACE information. Something like Quantity "ada"/Quantity "lovelace" that enables to talk with whatever "level" and always be correct....

Agreed, that would definitely be nice to have! Though luckily, we don't need something too complicated for this library, as:

1. We only have one coin type, which encodes a number of Lovelace.
2. We never have to deal with values that are smaller than 1 Lovelace.

src/library/Cardano/CoinSelection.hs
jonathanknowles: > same nitpicking as above ;-)

:laughing:

jonathanknowles:
@paweljakubas wrote:

> LGTM!. I am very fond of commit summary - it helped me to comprehend this no-so-small PR!

Thank-you for this feedback, I appreciate it. I tried very hard to write a summary that would make this PR easier to understand. I'm glad that you found it useful! :+1:

Issue #20, opened 2020-03-20
Reuse `cardano-coin-selection` library within `cardano-wallet`.


jonathanknowles:
# Context

When the work to create a separate [**coin selection and fee balancing library**](https://github.com/input-output-hk/cardano-wallet/milestone/58) is complete, we'd like to be able to reuse the library internally within the [`cardano-wallet`](https://github.com/input-output-hk/cardano-wallet) library.

# Decision

Make the [`cardano-wallet` library](https://github.com/input-output-hk/cardano-wallet) depend upon the Hackage version of [`cardano-coin-selection`](https://github.com/input-output-hk/cardano-coin-selection), and remove all local copies of coin selection code.

# Acceptance Criteria

- [ ] The [`cardano-wallet`](https://github.com/input-output-hk/cardano-wallet) library _must_ depend upon the **Hackage** version of [`cardano-coin-selection`](https://github.com/input-output-hk/cardano-coin-selection).
- [ ] The [`cardano-wallet`](https://github.com/input-output-hk/cardano-wallet) library _must not_ duplicate code that is included within the published [`cardano-coin-selection`](https://github.com/input-output-hk/cardano-coin-selection) library.

# Development 

_Section to be added._

# QA 

_Section to be added._

PR #44, opened 2020-04-15
Disable Automatic Algorithm Fallback

jonathanknowles:
## Related Issue

#30 

## Summary

The `cardano-wallet` library currently performs **automatic fallback** from `Random-Improve` to `Largest-First` if the former algorithm is unable to compute a valid coin selection.

However, this automatic fallback might not be appropriate for all users of the coin selection library.

This PR:

- [x] **removes** the automatic fallback logic within the implementation of `Random-Improve`.
- [x] **adds** an appropriate error handling function, since `Random-Improve` can no longer rely on the error handling provided by `Largest-First`.

The error handling is already tested by our existing test suite.

### DIFF
--- a/src/library/Cardano/CoinSelection/LargestFirst.hs
+++ b/src/library/Cardano/CoinSelection/LargestFirst.hs
@@ -201,7 +201,7 @@ payForOutputs options utxo outputsRequested =
       | otherwise =
           ErrMaximumInputCountExceeded inputCountMax
     amountAvailable =
-        fromIntegral $ unCoin $ coinMapValue utxo
+        unCoin $ coinMapValue utxo
     amountRequested =
         unCoin $ coinMapValue outputsRequested
     inputCountMax =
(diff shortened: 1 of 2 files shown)


jonathanknowles:
@paweljakubas wrote:

> if we disable largest first fallback, would be good to show in testing that we can recreate old setup. by invoking `random` and then rely on `largestFirst`. And here deliver `test` showing that we do not have fallback (there were test like that to my best knowledge in cardano-wallet), and when having the composition with fallback it service those cases.

Thanks for raising this point!

I agree, it's very important to preserve the existing behaviour of `cardano-wallet` when we eventually reuse `cardano-coin-selection` from `cardano-wallet`, replacing the old code.

I've added an item to the acceptance criteria for this task [here](https://github.com/input-output-hk/cardano-coin-selection/issues/20]), which highlights the need for testing to make sure that fallback continues to happen as expected.

Issue #17, opened 2020-02-26
Generalize types within `cardano-coin-selection` library.


jonathanknowles:
# Context

The [`cardano-coin-selection`](https://github.com/input-output-hk/cardano-coin-selection) library currently duplicates some of the types that are defined within the [`cardano-wallet`](https://github.com/input-output-hk/cardano-wallet/) library.

It would be desirable to replace these duplicated types with abstract types or interfaces, allowing consumers of the library (including `cardano-wallet`) to use their own concrete types.

# Decision

Revise the types used in the [`cardano-coin-selection`](https://github.com/input-output-hk/cardano-coin-selection) library, replacing concrete types with abstract types (or interfaces) where it makes sense.

## Types to Generalize

- [x] [`Cardano.CoinSelection.CoinSelectionLimit`](https://github.com/input-output-hk/cardano-coin-selection/blob/0dfd846c68669d60fdff4f77dc8ae19ae25bf0e1/src/library/Cardano/CoinSelection.hs#L278)
     This was originally defined in terms of `Word8`. It has been widened to `Word16`.

- [x] `TxIn` (used in the inputs of a `CoinSelection`)
     Since coin selection algorithms are agnostic to the type of input identifier, this has been replaced with the type variable `i`.

- [x] `Address` (used in the outputs of a `CoinSelection`)
     Since coin selection algorithms are agnostic to the type of output identifier, this has been replaced with a type variable `o`.

- [x] `UTxO`
    This has been replaced with `CoinMap i`, where `i` refers to an input identifier type.

- [x] `[TxOut]`
    This has been replaced with `CoinMap o`, where `o` refers to an output identifier type.

- [x] `Hash`
    This has been removed, as it was part of `TxIn` which has also been removed.

- [x] `Coin`
     This was originally defined in terms of `Word64`, and has now been redefined in terms of `Natural`.

- [x] `Fee`
     This was originally defined in terms of `Word64`, and has now been redefined in terms of `Coin` (which is defined in terms of `Natural`).

- [x] `DustThreshold`
     This was originally defined in terms of `Word64`, and has now been redefined in terms of `Coin` (which is defined in terms of `Natural`).

jonathanknowles:
[...]

PR #50, opened 2020-04-16
Unbounded Coin Values

jonathanknowles:
## Related Issue

#17 

## Summary

This PR redefines **`Coin`** as a `newtype` wrapper around `Natural`, which:

- Provides the following constructors:
    - `coinFromNatural :: Natural -> Coin`
    - `coinFromIntegral :: Integral i => i -> Maybe Coin`
- Provides a _safe_ `subtract` function (`Coin -> Coin -> Maybe Coin`).
- Provides a _safe_ `distance` function (`Coin -> Coin -> Coin`).
- Does _not_ provide a `Num` instance (as this requires the definition of an unsafe `-` operator).

Additionally, this PR:

- [x] Redefines `Fee` and `DustThreshold` in terms of `Coin`.
- [x] Updates the definition of `splitCoin` to remove the special processing of overflow values.
- [x] Redefines several functions to avoid subtraction (or use safe subtraction).
- [x] Updates `Arbitrary` instances for `Coin`, `Fee` and `DustThreshold`.
- [x] Adds extra coverage checks to some properties.
- [x] Removes several redundant properties that check for "valid" coins. (These are no longer necessary, as values of `Coin` can now no longer be invalid by construction.)


rvl:
Seems OK. Hard to really know what the maximum supply of any arbitrary token might be.
Although I'm pretty sure all of them would exceed the range of Int64 or Word64.
I noticed that we are going from an unsigned to signed coin type. Any reason for that?

jonathanknowles:
@rvl wrote:

> I noticed that we are going from an unsigned to signed coin type. Any reason for that?

This is a great point :+1:. I'm currently testing out the use of `Natural` rather than `Integer`.

I'll update the PR after I can get the build to work, and the test suite to pass.

jonathanknowles:
@rvl wrote:

> > I noticed that we are going from an unsigned to signed coin type. Any reason for that?

@jonathanknowles responded:
 
> This is a great point +1. I'm currently testing out the use of `Natural` rather than `Integer`.
> I'll update the PR after I can get the build to work, and the test suite to pass.

**Results of preliminary investigation:**

Swapping out `Integer` for `Natural` causes around 28 test failures due to `ArithmeticUnderflow`.

The `ArithmeticUnderflow` error happens when a code path attempts to create a negative value of `Natural` at run-time.

The reason we never saw this with `Word64` is that if an operation attempts to create a negative value of `Word64`, the value is silently wrapped around (i.e., all arithmetic is performed modulo 2^64).

This obviously requires some further investigation, as it indicates that our test suite is relying on this wrap-around behaviour in several places in order to pass. But I can't really conceive of a situation where such behaviour is desirable.

paweljakubas:
If `Word64` is too restrictive for us because of upper bound, then I would go for `Natural` (to have sound lower bound). If we want to `eliminate` upper bound. If we want to maintain it then I would go for something like `CoinDomain a` with lowerBound=0 and upperBound established by `a` or lower than the one following from `a` if needed. And throughout the code would use lowerBound and upperBound of this new type.

jonathanknowles:
@paweljakubas wrote:

> If `Word64` is too restrictive for us because of upper bound, then I would go for `Natural` (to have sound lower bound). If we want to `eliminate` upper bound. If we want to maintain it then I would go for something like `CoinDomain a` with lowerBound=0 and upperBound established by `a` or lower than the one following from `a` if needed. And throughout the code would use lowerBound and upperBound of this new type.

Thanks for your review of this!

One of the stated goals of this PR is to **remove the upper bound** on **`Coin`**, so that we no longer need to deal with overflow in functions like `splitCoin`. In this sense, the PR does succeed, as we have indeed removed all logic that deals with upper bounds.

As for the choice of _underlying representation_ for `Coin`, I agree that it is ideal if we could eventually move to `Natural`, and completely forbid the creation of negative values, even if they are just temporary values to be thrown away later on in a calculation.

However, I propose that we tackle this in two stages:

1. Change from `Word64` to `Integer` (this PR).
2. Change from `Integer` to `Natural` (a future PR).

The justification for splitting this work up is that there are actually quite a few places in the code where we perform subtraction (potentially generating negative values). Restructuring the code to avoid subtraction completely (or at least compartmentalize it behind carefully tested functions) will likely require a bit of work to get right.

It's worth pointing out that:
* Before applying this PR, we did allow the temporary generation of negative `Coin` values, but these were represented as silently-wrapped-around `Word64` values.
* If we apply this PR (and do nothing else), we will continue to allow the temporary generation of negative `Coin` values, but these will now be represented as negative `Integer` values.

Thoughts?

jonathanknowles:
> **Stage 1**: Change from `Word64` to `Integer` (this PR).
> **Stage 2**: Change from `Integer` to `Natural` (a future PR).

Just for the record:

After an internal discussion, it was decided that:

1. We will go **directly** from `Word64` to `Natural` and resolve all test suite errors in **a single PR**, without merging intermediate PRs that use `Integer`.
2. To make this possible, we will systematically eliminate subtraction operations involving `Coin` and `Fee` values from the code base.
3. In cases where it's impossible or impractical to eliminate subtraction, we'll ensure that it's guarded appropriately to avoid any run-time errors (using safe function wrappers that return `Maybe` or other similar types).

Therefore, I'll move this PR back to the WIP state for the moment.

src/internal/Internal/SafeNatural.hs
```diff
@@ -0,0 +1,130 @@
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+module Internal.SafeNatural
```
KtorZ, line 6: Isn't the point of `Coin` to be the `SafeNatural` :grimacing: ? We may be over-doing it here.

jonathanknowles: @KtorZ wrote:
> Isn't the point of `Coin` to be the `SafeNatural` grimacing ?

That's a good question. `SafeNatural` is actually used as the underlying representation of two types:

- `Coin`
- `DustThreshold`

This is partly because it doesn't make sense for a dust threshold to be negative, and partly for convenience, as much of the time we are comparing dust thresholds with coin values, and it helps to avoid unnecessary conversions.

Since the `SafeNatural` type is encapsulated in its own module, and since that module only exports smart constructors and total functions on `SafeNatural`, it should be relatively hard to accidentally create negative values of `Coin`, due to the protection conferred by the underlying type.

src/library/Cardano/CoinSelection.hs
```diff
@@ -222,13 +192,13 @@ data CoinSelectionOptions i o e = CoinSelectionOptions
...
 data CoinSelectionError e
-    = ErrUtxoBalanceInsufficient Word64 Word64
+    = ErrUtxoBalanceInsufficient Coin Coin
     -- ^ The UTxO balance was insufficient to cover the total payment amount.
     --
     -- Records the /UTxO balance/, as well as the /total value/ of the payment
     -- we tried to make.
     --
-    | ErrUtxoNotFragmentedEnough Word64 Word64
+    | ErrUtxoNotFragmentedEnough Integer Integer
```
paweljakubas, line 201: `Integer` is here better than Word64? Why not `Natural` if `Word64` is too restrictive ? :thinking:

src/library/Cardano/CoinSelection.hs
```diff
@@ -240,7 +210,7 @@ data CoinSelectionError e
     -- available UTxO entries were depleted before all the requested
     -- transaction outputs could be paid for.
     --
-    | ErrMaximumInputCountExceeded Word64
+    | ErrMaximumInputCountExceeded Integer
```
paweljakubas, line 213: the same remark/question like above?

src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -100,32 +105,21 @@ import qualified Data.List.NonEmpty as NE
 -- This type is isomorphic to 'Coin'.
 --
 newtype Fee = Fee
-    { unFee :: Word64 }
+    { unFee :: Integer }
```
paweljakubas, line 108: `Fee` can be negative? :thinking:

jonathanknowles: > `Fee` can be negative?

In order to limit the scope of this PR, I have not (yet) attempted to restrict `Fee` in the same way as `Coin` and `DustThreshold`.

If we can determine that our codebase never creates any intermediate negative values of `Fee`, then we could also redefine `Fee` in terms of `SafeNatural`.

src/library/Cardano/CoinSelection.hs
jonathanknowles: Well spotted. I've changed this (and other similar values) to use `Natural`.

Fixed in cc259b15b77499ce80a101608b44f55471b830b0.

src/library/Cardano/CoinSelection.hs
jonathanknowles: Well spotted. I've changed this (and other similar values) to use `Natural`.

Fixed in cc259b15b77499ce80a101608b44f55471b830b0.

src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -488,65 +488,41 @@ remainingFee FeeEstimator {estimateFee} s
...
+            Coin <$> zipWith SN.add (unCoin <$> coinsToIncrease) increments
+          where
+            increments = zipWith SN.add majorIncrements minorIncrements
+            majorIncrements = repeat increment
+            minorIncrements = replicate (SN.toIntegral shortfall) SN.one
+                <> repeat SN.zero
+        _ | SN.isPositive coinToSplit ->
+            [Coin coinToSplit]
+        _ ->
+            []
```
jonathanknowles, line 518: This is a fresh implementation that avoids subtraction.

Fortunately, the function is well covered by several property tests.

src/test/Cardano/CoinSelection/TypesSpec.hs
```diff
@@ -45,11 +48,10 @@ import Test.QuickCheck
 
 import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
+import qualified Internal.SafeNatural as SN
 
 spec :: Spec
 spec = do
-    describe "Generators are valid" $ do
-        it "Arbitrary Coin" $ property coinIsValid
```
jonathanknowles, line 52: This property (and several others like it) are no longer necessary.

src/internal/Internal/SafeNatural.hs
KtorZ: `DustThreshold` used to be expressed in `Coin`, so by definition couldn't possibly be negative. I still fail to see what value do we gain by having this extra indirection to `SafeNatural` which requires redefining all the internals of an already quite idiomatic type of the Haskell eco-system: `Natural` :( .

src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -305,13 +300,14 @@ reduceChangeOutputs threshold (Fee totalFee) changeOutputs
             [] -> []
   where
     payFee :: (Fee, Coin) -> Coin
-    payFee (Fee f, Coin c) = Coin (c - f)
```
KtorZ, line 308: I am not sure how we ended up with this function in the first place. This has probably slipped through review and refactoring. On the original code, this used to be:

```hs
-- | Reduce single change output by a given fee amount. If fees are too big for
-- a single coin, returns a `Coin 0`.
reduceSingleChange :: (Fee, Coin) -> Coin
reduceSingleChange (Fee fee, Coin chng)
    | chng >= fee =
          Coin (chng - fee)
    | otherwise =
          Coin 0
```

And the case of underflow was actually properly handled. It now is anyway, but it still feels a lot heavier than the guard above and the use a mere `Natural`.

src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -305,13 +300,14 @@ reduceChangeOutputs threshold (Fee totalFee) changeOutputs
...
+    payFee (Fee f, Coin c) = Coin $
+        fromMaybe SN.zero ((c `SN.sub`) =<< SN.fromIntegral f)
 
     positiveChangeOutputs :: [Coin]
-    positiveChangeOutputs = filter (> Coin 0) changeOutputs
+    positiveChangeOutputs =
+        Coin <$> filter SN.isPositive (unCoin <$> changeOutputs)
```
KtorZ, line 308: I find the term `Positive` actually a lot less clearer than a mathematical inequality `> 0`. The reason for this is that maths are unambiguous (in this context at least) whereas 'Positive' requires an careful understanding of the English language. why I am pointing that out here is because, while `Positive` in English means `> 0`, `Positive` in French means `>= 0` and it's probably the case in other languages too. 

When possible, sticking to mathematical notation may be better.

src/library/Cardano/CoinSelection/Fee.hs
jonathanknowles: @paweljakubas Fixed in 79fdc177d677ad1974ef426b836a0e644fd4a25e.

src/library/Cardano/CoinSelection/Fee.hs
jonathanknowles: > I find the term `Positive` actually a lot less clearer than a mathematical inequality `> 0`. The reason for this is that maths are unambiguous (in this context at least) whereas 'Positive' requires an careful understanding of the English language.

Thanks. I wasn't aware that this distinction existed in French. After researching this, I discovered the term "strictement positif", which seems similar to the distinction between a subset and a "strict" subset.

For the avoidance of any doubt, I've replaced these with `> SN.zero`.

Fixed in e98fd62a95b743dc4b821c2be581861ee50d0978.

src/library/Cardano/CoinSelection/Fee.hs
jonathanknowles: @KtorZ wrote:
> I am not sure how we ended up with this function in the first place. This has probably slipped through review and refactoring.

Quite likely.

Though to my mind, this really does support the case for having a type that _doesn't_ support unchecked subtraction.

If the underlying type implements `Num` (as `Natural` does), it becomes easy for a future programmer to _unintentionally_ introduce a regression just by using `-` in an unchecked context.

At least with `SafeNatural`, the attack surface is greatly reduced, as now we only have:

```hs
sub      :: SafeNatural -> SafeNatural -> Maybe SafeNatural
distance :: SafeNatural -> SafeNatural ->       SafeNatural
```

In my opinion, being able to rely on the type system in this way is _much_ safer than relying on code review to spot the use of an unchecked subtraction. (Of course, this doesn't negate the need for good code review.)

> it still feels a lot heavier than the guard above and the use a mere Natural.

After change 79fdc177d677ad1974ef426b836a0e644fd4a25e, this is now similar in spirit to the original:
```hs
    payFee :: (Fee, Coin) -> Coin                                                                                                                                                     
    payFee (Fee f, Coin c) = Coin $                                                                                                                                                   
        fromMaybe SN.zero (c `SN.sub` f)   
```
Which should be similar (in spirit) to the original.

src/internal/Internal/SafeNatural.hs
jonathanknowles: @KtorZ wrote:
> I still fail to see what value do we gain by having this extra indirection to SafeNatural which requires redefining all the internals of an already quite idiomatic type of the Haskell eco-system: Natural

The issue with Natural is that it provides a `Num` instance, which requires an implementation of the `-` operator:
```hs
(-) :: a -> a -> a
```

Just having this operator available means that it's trivial for a programmer (several months down the line) to accidentally introduce an **_unchecked_** subtraction back into the code.

Of course, good code review and good test coverage should ideally stop this from happening, but sometimes it doesn't. (See https://github.com/input-output-hk/cardano-coin-selection/pull/50#discussion_r411138543 for a great example.)

KtorZ: I get your point but I still disagree. I think it's pushing the complexity line a bit too far for only little added benefits. This isn't code that is changing _much_ and, there are all sort of manipulation with the fee and coins to be careful with already. So it's not like we take substraction and its consequences out of consideration in this code. 

I am quite curious though about the added complexity pushed onto the library users. If the added complexity is only visible to us internally then.. whatever. But if this results in an extra cost for library users, then it's a no go.

jonathanknowles: @KtorZ wrote:

> I am quite curious though about the added complexity pushed onto the library users. If the added complexity is only visible to us internally then.. whatever. But if this results in an extra cost for library users, then it's a no go.

I don't think that a user will see much complexity, as it's deliberately hidden from view.

For converting a coin to some other type (after running a coin selection algorithm) we provide:
```hs
coinToIntegral :: Integral i -> Coin -> i
```
For constructing values of type `Coin` (in order to pass in to a coin selection algorithm), we provide a smart constructor:
```hs
coin :: Integral i -> Maybe Coin
```
We **_could_** make this even easier by giving a choice of functions:
```hs
coinFromIntegral :: Integral i -> Maybe Coin
coinFromNatural  :: Natural    ->       Coin
```
Thoughts?

paweljakubas: why not to encapsulate everything in `Coin` and then use it inside `Fee` and `DustThreshold` ? (ie., resign from `SafeNatural` in a sense that `Coin` would have it, and then add `Num` instance for `Coin` :thinking:

jonathanknowles: @paweljakubas wrote:
> then add Num instance for Coin

If we provide a `Num` instance for `Coin`, then it opens the door to creating coins with negative values. This is because the `Num` class requires an implementation of `-`, which has the following type:

```hs
(-) :: a -> a -> a
```

The only way to provide a definition that satisfies this type without allowing negative coin values is to throw a run-time exception if the second argument is greater than the first.

But the `SafeNatural` type makes this **impossible** by _not_ providing a `Num` instance, instead providing:

```hs
sub      :: SafeNatural -> SafeNatural -> Maybe SafeNatural
distance :: SafeNatural -> SafeNatural ->       SafeNatural
```

Of course, we _could_ merge the definition of `Coin` and `SafeNatural`, and then redefine `Fee` and `DustThreshold` in terms of `Coin`, but this really wouldn't change much from the user's point of view, if at all.

paweljakubas: and in case of impl `(-)` of `Num` we cannot have `assert` to outlaw negative values? We have instances of `Num` for `GHC.Natural` implemented, like here https://hackage.haskell.org/package/base-4.12.0.0/docs/src/GHC.Num.html#line-143 We can maybe have in case of `(-)` some customized mechanism or fallback on what `Natural` offers? No?

jonathanknowles: @paweljakubas wrote:
> why not to encapsulate everything in Coin and then use it inside Fee and DustThreshold 

Fixed in 01a3a62160221f5ab8adf98d6e11de1fac611147.

jonathanknowles: @KtorZ The public API now exposes the following constructor (and deconstructor):
```hs
coinFromNatural :: Natural -> Coin

coinToNatural :: Coin -> Natural
```
Hopefully this should be fairly easy to use, but let me know if you'd prefer a different way.

(Internally, we also have `coinFromIntegral`, but this isn't exposed in the public API.)

jonathanknowles:
Just for the record:

The latest update (01a3a62160221f5ab8adf98d6e11de1fac611147) coalesces `Coin` with `SafeNatural`, and redefines `Fee` and `DustThreshold` in terms of `Coin`.

src/internal/Internal/SafeNatural.hs
jonathanknowles: @KtorZ Additionally, the definition of `SafeNatural` is now merged into `Coin`.

`Coin` is now a `newtype` wrapper around `Natural`. We retain the set of safe operations, as before.

`Fee` and `DustThreshold` are now just `newtype` wrappers around `Coin`.

PR #53, opened 2020-04-21
Public API of `CoinSelection` Module.

jonathanknowles:
## Related Issue

#18 
#30 

## Summary

This PR makes a number of changes aimed at improving the public API and documentation:

- [x] Revises the overall structure of the documentation for the main `CoinSelection` module.
- [x] Uses `OPTIONS_HADDOCK prune` to hide internal functions from Haddock documentation.
- [x] Gives each error condition its own separate type. Advantages:
    - [x] Parameters to errors can be named, making `Show` instances more informative.
    - [x] The documentation for each error can be more informative, as it can be separated into paragraphs (rather than squashed into a table cell).
- [x] Removes user-defined validation. If a caller wishes to validate a generated `CoinSelection`, they can easily validate it themselves after calling `selectCoins`.

### DIFF
--- a/src/library/Cardano/CoinSelection.hs
+++ b/src/library/Cardano/CoinSelection.hs
@@ -297,7 +297,7 @@ data CoinSelectionError
     -- available UTxO entries were depleted before all the requested
     -- transaction outputs could be paid for.
     --
-    | ErrMaximumInputCountExceeded Natural
+    | ErrLimitExceeded Natural
     -- ^ The number of UTxO entries needed to cover the requested payment
     -- exceeded the upper limit specified by 'limit'.
     --
--- a/src/library/Cardano/CoinSelection.hs
+++ b/src/library/Cardano/CoinSelection.hs
@@ -12,7 +12,7 @@
 -- Provides general functions and types relating to coin selection.
 --
 -- The 'CoinSelection' type represents a __coin selection__, the basis for a
--- /transaction/.
+-- /transaction/ in a UTxO-based blockchain.
 --
 -- The 'CoinSelectionAlgorithm' type provides a __common interface__ to
 -- algorithms that generate coin selections.
--- a/src/library/Cardano/CoinSelection.hs
+++ b/src/library/Cardano/CoinSelection.hs
@@ -39,6 +39,7 @@ module Cardano.CoinSelection
 
       -- * Coin Selection Algorithms
     , CoinSelectionAlgorithm (..)
+    , CoinSelectionParameters (..)
     , CoinSelectionInputLimit (..)
     , CoinSelectionError (..)
 
@@ -165,11 +166,21 @@ coinMapValue = mconcat . fmap entryValue . coinMapToList
 --
 newtype CoinSelectionAlgorithm i o m = CoinSelectionAlgorithm
     { selectCoins
-        :: CoinSelectionInputLimit
-        -> CoinMap i
-        -> CoinMap o
+        :: CoinSelectionParameters i o
         -> ExceptT CoinSelectionError m (CoinSelection i o, CoinMap i)
     }
+
+-- | The complete set of parameters required for a 'CoinSelectionAlgorithm'.
+--
+data CoinSelectionParameters i o = CoinSelectionParameters
+    { inputLimit :: CoinSelectionInputLimit
+        -- ^ A limit on the number of inputs that can be selected.
+    , inputsAvailable :: CoinMap i
+        -- ^ The set of inputs available for selection.
+    , outputsRequested :: CoinMap o
+        -- ^ The set of outputs requested for payment.
+    }
+    deriving Generic
 
 -- | Represents the __result__ of running a coin selection algorithm.
 --
--- a/src/library/Cardano/CoinSelection.hs
+++ b/src/library/Cardano/CoinSelection.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE RankNTypes #-}
+{-# OPTIONS_HADDOCK prune #-}
 
 -- |
 -- Copyright: © 2018-2020 IOHK
@@ -29,7 +30,6 @@ module Cardano.CoinSelection
     , coinMapFromList
     , coinMapToList
     , coinMapValue
-    , coinMapRandomEntry
 
       -- * Coin Selection
     , CoinSelection (..)
@@ -41,6 +41,9 @@ module Cardano.CoinSelection
     , CoinSelectionAlgorithm (..)
     , CoinSelectionOptions (..)
     , CoinSelectionError (..)
+
+      -- # Internal Functions
+    , coinMapRandomEntry
 
     ) where
 
@@ -135,24 +138,6 @@ coinMapToList = fmap (uncurry CoinMapEntry) . Map.toList . unCoinMap
 --
 coinMapValue :: CoinMap a -> Coin
 coinMapValue = mconcat . fmap entryValue . coinMapToList
-
--- | Selects an entry at random from a 'CoinMap', returning both the selected
---   entry and the map with the entry removed.
---
--- If the given map is empty, this function returns 'Nothing'.
---
-coinMapRandomEntry
-    :: MonadRandom m
-    => CoinMap a
-    -> m (Maybe (CoinMapEntry a), CoinMap a)
-coinMapRandomEntry (CoinMap m)
-    | Map.null m =
-        return (Nothing, CoinMap m)
-    | otherwise = do
-        ix <- fromEnum <$> generateBetween 0 (toEnum (Map.size m - 1))
-        let entry = uncurry CoinMapEntry $ Map.elemAt ix m
-        let remainder = CoinMap $ Map.deleteAt ix m
-        return (Just entry, remainder)
 
 --------------------------------------------------------------------------------
 -- Coin Selection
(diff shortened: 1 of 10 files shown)


src/library/Cardano/CoinSelection.hs
```diff
@@ -3,44 +3,55 @@
...
 
 -- |
 -- Copyright: © 2018-2020 IOHK
 -- License: Apache-2.0
 --
 -- Provides general functions and types relating to coin selection.
 --
--- The 'CoinSelectionAlgorithm' type provides a /common interface/ to coin
--- selection algorithms.
+-- The 'CoinSelection' type represents a __coin selection__, the basis for a
```
rvl, line 14: If this is going on Haddock I think it's worth noting that this is for UTxO-based blockchains such as Bitcoin and Cardano.

src/library/Cardano/CoinSelection.hs
```diff
@@ -136,71 +147,105 @@ coinMapToList = fmap (uncurry CoinMapEntry) . Map.toList . unCoinMap
...
 
--- | Provides a common interface for coin selection algorithms.
+-- | Provides a __common interface__ for coin selection algorithms.
+--
+-- The function 'selectCoins', when applied to the given
+-- 'CoinSelectionParameters' object (with /available inputs/ and /requested/
+-- /outputs/), will generate a 'CoinSelectionResult' (with /remaining inputs/
+-- and a /coin selection/).
+--
+newtype CoinSelectionAlgorithm i o m = CoinSelectionAlgorithm
```
rvl, line 161: Can't this be a type alias instead?

src/library/Cardano/CoinSelection.hs
```diff
@@ -226,51 +271,98 @@ sumOutputs =  coinMapValue . outputs
...
 
--- | Represents a set of options to be passed to a coin selection algorithm.
+-- | Defines an __inclusive upper bound__ on the /number/ of inputs that
+--   a 'CoinSelectionAlgorithm' is allowed to select.
 --
-data CoinSelectionOptions i o e = CoinSelectionOptions
-    { maximumInputCount
+newtype CoinSelectionLimit = CoinSelectionLimit
+    { calculateLimit
         :: Word8 -> Word8
```
rvl, line 279: It occurred to me that some blockchains might permit more than 255 tx inputs.
But that's for another PR I suppose.

src/library/Cardano/CoinSelection.hs
```diff
@@ -226,51 +271,98 @@ sumOutputs =  coinMapValue . outputs
...
+--
+data CoinSelectionError
+    = InputValueInsufficient
+        InputValueInsufficientError
+    | InputCountInsufficient
+        InputCountInsufficientError
+    | InputLimitExceeded
+        InputLimitExceededError
+    | InputsExhausted
+        InputsExhaustedError
```
rvl, line 297: This does seem to make the Haddocks easier to read.

src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -9,36 +9,32 @@
...
-    , FeeOptions (..)
-    , ErrAdjustForFee(..)
     , adjustForFee
-    , reduceChangeOutputs
+    , FeeOptions (..)
+    , FeeAdjustmentError (..)
 
-      -- * Dust Processing
-    , DustThreshold (..)
+      -- # Internal Functions
```
rvl, line 33: This doesn't need the `#` does it?

src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -9,36 +9,32 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# OPTIONS_HADDOCK prune #-}
```
rvl, line 12: You know, I don't think it would be that bad to show internal functions in haddock, under an internal functions heading.

Users are probably interested in the precise details of how it is implemented. Because they might need to decide between using one of these algorithms or making their own.

src/library/Cardano/CoinSelection/LargestFirst.hs
```diff
@@ -154,73 +155,77 @@ import qualified Internal.Coin as C
...
-    -> ExceptT (CoinSelectionError e) m (CoinSelection i o, CoinMap i)
-payForOutputs options utxo outputsRequested =
+    => CoinSelectionParameters i o
+    -> ExceptT CoinSelectionError m (CoinSelectionResult i o)
+payForOutputs params =
     case foldM payForOutput (utxoDescending, mempty) outputsDescending of
         Just (utxoRemaining, selection) ->
-            validateSelection selection $>
-                (selection, coinMapFromList utxoRemaining)
+            pure $ CoinSelectionResult selection $ coinMapFromList utxoRemaining
```
rvl, line 191: Makes sense.

src/library/Cardano/CoinSelection.hs
jonathanknowles: > If this is going on Haddock I think it's worth noting that this is for UTxO-based blockchains such as Bitcoin and Cardano.

Good point! Fixed in ea8568fba6669cb5e64a3fb17f1966a7d8fa7280.

src/library/Cardano/CoinSelection.hs
jonathanknowles: > It occurred to me that some blockchains might permit more than 255 tx inputs.

Great point!

I would also like to change this, though the `idealBatchSize` function from `Migration` currently depends on this type. If that function can be changed to work with a wider type (in a future PR), then we can raise this limit.

src/library/Cardano/CoinSelection.hs
jonathanknowles: Yes. It seems Haddock doesn't cope well with paragraphs within error constructors. (As the text is squashed within a table cell.)

Another benefit of giving each error a separate type is that we can also give names to error parameters, which should help with debugging.

src/library/Cardano/CoinSelection/Fee.hs
jonathanknowles: > This doesn't need the `#` does it?

Strictly speaking, it's not necessary.

I just put it there to create a "section title" that isn't picked up by Haddock.

src/library/Cardano/CoinSelection/Fee.hs
paweljakubas: I agree here

src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -181,28 +166,40 @@ adjustForFee
...
     nullFee opt = estimateFee (feeEstimator opt) coinSel == Fee C.zero
 
--- | The sender pays fee in this scenario, so fees are removed from the change
+--------------------------------------------------------------------------------
+-- Internal Functions
+--------------------------------------------------------------------------------
+
+-- Calculates the current fee associated with a given 'CoinSelection'.
+--
+-- If the result would be less than zero, returns 'Nothing'.
```
paweljakubas, line 183: You mean: 
`If the result is less than zero, then 'Nothing' is returned.`
?

src/library/Cardano/CoinSelection/Fee.hs
jonathanknowles: > You know, I don't think it would be that bad to show internal functions in haddock, under an internal functions heading.

I definitely see your point that users might wish to see more internal details. Currently, they can still import the identifier into their own code, so there's no hard restriction.

The rationale behind doing it this way is explained here:
https://github.com/input-output-hk/cardano-coin-selection/pull/40#issuecomment-611970563

We can also slowly "unhide" parts of the API as we go forward in time.

src/library/Cardano/CoinSelection/Migration.hs
```diff
@@ -34,8 +35,13 @@
 -- that a wallet is not "fragmented enough").
 
 module Cardano.CoinSelection.Migration
-    ( depleteUTxO
+    (
+      -- * Coin Selection for Migration
+      selectCoins
+
+      -- # Internal Functions
```
paweljakubas, line 42: not `*` ?

src/library/Cardano/CoinSelection/Fee.hs
jonathanknowles: Thanks for the feedback on this comment.

Actually, either phrasing would work, since the "would" here expresses a hypothetical (the _returned_ result can never be less than zero).

Though I'll change it to your suggested wording, as I agree it's shorter and clearer! :+1: 

Fixed in 8e42eca4e1612b50f41a553f7bb6fa8fa97dc7a8.

src/library/Cardano/CoinSelection/Migration.hs
jonathanknowles: Using `#` here means that it is not processed by Haddock. (Users can still import the functions, but they're not described in the generated Haddock documentation.)

src/library/Cardano/CoinSelection.hs
jonathanknowles: > Can't this be a type alias instead?

It could be. Though would there be an advantage to making it an alias?

I'll leave that for a future PR though. I've added an item here: https://github.com/input-output-hk/cardano-coin-selection/issues/30

jonathanknowles:
@paweljakubas wrote:

> very nice documentation pimping;-) great job!

Thanks for reviewing it!

PR #55, opened 2020-04-22
Revise Documentation for the `Migration` Module.

KtorZ:
## Related Issue

#18 

## Summary

This PR:

- [x] Rewrites the documentation for the `Migration` module to be more general, removing references to Cardano-specific blockchain values.
- [x] Introduces type `BatchSize`, which makes the batch size argument easier to refer to in documentation, and provides a way for users to relate the `idealBatchSize` and `selectCoins` functions to one another.
- [x] Exposes `idealBatchSize` in the public API.

In addition, this PR:

- [x] Adds the `@since 1.0.0` metadata attribute to each symbol exposed in the public API. (The necessity of this was discussed in an earlier review, and a consensus was reached that we will include an attribute for each symbol in the public API, right from the first release.)


src/library/Cardano/CoinSelection/Migration.hs
```diff
@@ -8,11 +8,10 @@
 -- Copyright: © 2018-2020 IOHK
 -- License: Apache-2.0
 --
--- This module contains an algorithm to select coins for migration from legacy
--- wallets to newer wallets.
+-- This module contains an algorithm to select coins for emptying a wallet.
```
jonathanknowles, line 11: I think it would be better to use the word "migration" here, as "emptying" has the possible connotation that funds would be thrown away (when in fact, we want them to be _migrated_ _somewhere else_).

How about:
```suggestion
-- This module contains an algorithm for migrating all funds away from a wallet.
```

src/library/Cardano/CoinSelection/Migration.hs
```diff
@@ -8,11 +8,10 @@
...
 --
--- We want users to be able to migrate their funds from a legacy random wallet
--- to a new sequential wallet. To do this, we have to move funds from a wallet
+-- A typical use-case could be users who want to migrate their funds from an old
+-- wallet to a new wallet. To do this, we have to move funds from a wallet
```
jonathanknowles, line 14: :+1:

src/library/Cardano/CoinSelection/Migration.hs
```diff
@@ -8,11 +8,10 @@
...
 -- to another by making transactions. Funds are ultimately a sum of many coins
 -- (a.k.a UTxOs). In a transaction, we can select a few coins, and send them to
 -- addresses, effectively creating new coins / UTxOs doing this.
```
jonathanknowles, line 17: Perhaps the last couple of lines are superfluous, as they pertain to transactions in general, rather than specifically to migration. (I think we already describe transactions elsewhere.)

How about:
```suggestion
-- to another by making a /series/ of transactions.
```

src/library/Cardano/CoinSelection/Migration.hs
```diff
@@ -184,12 +182,10 @@ selectCoins feeOpts batchSize utxo =
         put rest
         pure batch
 
---------------------------------------------------------------------------------
--- Internal Functions
---------------------------------------------------------------------------------
-
--- Try to find a fixed "ideal" number of input transactions that would generate
+-- | Try to find a fixed "ideal" number of input transactions that would generate
 -- relatively balanced transactions.
```
jonathanknowles, line 186: I have a couple of suggestions here. We could:

1. Add a definition of what it means to for a transaction to be "relatively-balanced"? (Justification: A new library user might not even know what a "balanced" transaction is.)

2. Add an explanation of why a user might want to use this function. (Of course, in the wallet, we use it to find a batch size that we then pass to the migration coin selection function, but perhaps this isn't very obvious to the user.)

src/library/Cardano/CoinSelection/Migration.hs
```diff
@@ -184,12 +182,10 @@ selectCoins feeOpts batchSize utxo =
...
+--
+-- @since 1.0.0
 idealBatchSize :: CoinSelectionLimit -> Word8
```
jonathanknowles, line 189: Should this `Word8` type be widened?

Related comment: https://github.com/input-output-hk/cardano-coin-selection/issues/17#issuecomment-618161639

jonathanknowles:
@KtorZ I suspect this part could also be revised, as the values seem very specific to Cardano:

https://github.com/input-output-hk/cardano-coin-selection/blob/f62d069062f196d3b251ac6e7a039a0030d929d6/src/library/Cardano/CoinSelection/Migration.hs#L19-L21

KtorZ:
@rvl At this stage it does not. But it'll be only useful by the time we make a second release indeed.

src/library/Cardano/CoinSelection/Migration.hs
KtorZ: So long as the `CoinSelectionLimit` remains unchanged, no, since the domain here is a direct consequence of the co-domain. Word8 is arguable small, and I do not imagine a transaction with more than 65535 inputs/ouputs! That would be insane. So if Word16 is an easy win, it could be an option.

jonathanknowles: `Word16` is indeed an easy win. I've created a PR from the above branch here: https://github.com/input-output-hk/cardano-coin-selection/pull/58

jonathanknowles:
@rvl wrote:

> Does putting "since 1.0.0" on every single symbol add anything useful?

After an internal team discussion, it was decided to include this attribute on every symbol exported in the public API right from the first published version.

jonathanknowles:
@jonathanknowles wrote:
> I suspect this part could also be revised, as the values seem very specific to Cardano:
> 
> https://github.com/input-output-hk/cardano-coin-selection/blob/f62d069062f196d3b251ac6e7a039a0030d929d6/src/library/Cardano/CoinSelection/Migration.hs#L19-L21

Fixed in 0bab2e888fa68d11c851642b95cc7742d6f54920.

src/library/Cardano/CoinSelection/Algorithm/Migration.hs
```diff
@@ -222,14 +216,18 @@ selectCoins FeeOptions{dustThreshold,feeEstimator,feeBalancingPolicy} batchSize
         put rest
         pure batch
 
---------------------------------------------------------------------------------
--- Internal Functions
---------------------------------------------------------------------------------
+-- | An upper limit for the number of 'inputs' to include in each coin selection
+--   generated by `selectCoins`.
```
jonathanknowles, line 220: This should probably be:

```suggestion
--   generated by 'selectCoins'.
```

src/library/Cardano/CoinSelection/Algorithm/Migration.hs
```diff
@@ -82,28 +62,42 @@ import qualified Internal.Coin as C
...
--- | Construct a list of coin selections / transactions to transfer the totality
--- of a user's wallet. The resulting 'CoinSelection' do not contain any
--- 'outputs', but only change coins (so there's no restriction about how
--- addresses are generated).
+-- | Creates a __series__ of coin selections that, when published as
+--   transactions, will have the effect of migrating all funds from one
+--   wallet to another.
+--
+-- Since UTxO-based blockchains typically impose limits on the sizes of
+-- individual transactions, and since individual UTxO sets can contain
```
paweljakubas, line 70: `UTxO sets` ot `UTxO set` ?

jonathanknowles: > `UTxO sets` ot `UTxO set` ?

This is correct, because we've written:

> "individual UTxO sets can contain arbitrarily many entries."

(The sentence is referring to the concept of UTxO sets in general.)

We could also have written:

> "an individual UTxO set can contain arbitrarily many entries."

(Either would be correct.)

PR #56, opened 2020-04-22
ignore automatically derived instances and record accessors in coverage

KtorZ:
These things are just creating noise in the coverage reports. With some overlay, we can better target which part of the code needs coverage. With this, we get to:

badge

### DIFF
--- a/.coverage/template.overlay
+++ b/.coverage/template.overlay
@@ -1,0 +1,38 @@
+module "Internal.Coin" {
+  tick "getSum" [Record accessor];
+  tick "unCoin" [Record accessor];
+  tick function "==" [derived Eq instance];
+  tick function "compare" [derived Ord instance];
+  tick function "showsPrec" [derived Show instance];
+}
+
+module "Internal.Invariant" {
+  inside "invariant" {
+    tick "error msg" [Never executed invariant];
+  }
+}
+
+module "Internal.Rounding" {
+  tick function "==" [derived Eq instance];
+  tick function "showsPrec" [derived Show instance];
+}
+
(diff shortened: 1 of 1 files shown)


.coverage/template.overlay
```diff
@@ -0,0 +1,38 @@
...
+  tick "getSum" [Record accessor];
+  tick "unCoin" [Record accessor];
+  tick function "==" [derived Eq instance];
+  tick function "compare" [derived Ord instance];
+  tick function "showsPrec" [derived Show instance];
+}
+
+module "Internal.Invariant" {
+  inside "invariant" {
+    tick "error msg" [Never executed invariant];
```
jonathanknowles, line 11: This reminded me that the invariants for `adjustForFee` and `distributeFee` are never tested.

I've added these to the list on https://github.com/input-output-hk/cardano-coin-selection/issues/22.

KtorZ: Yes, I actually realized yesterday that the changes on the fee balancing from cardano-wallet hasn't been ported to this library. I did ping you on the corresponding PR but I believe it got lost in translation. The PR actually _removes_ the invariant and, adds a quite extensive property for testing the balancing. It also enables two balancing mode, the "cheap" one that might create slightly unbalanced transaction if allowed by the node, and the "pay and continue" one that requires picking extra inputs to cover dangling output if any. 

The only drawback so far is that, we loose the "dust property" in the fee balancing. So, as a result of balancing for fee, the algorithm might actually create dust. It should be possible to work around that I think.

See: https://github.com/input-output-hk/cardano-wallet/pull/1578

PR #59, opened 2020-04-23
Improve Documentation for `Fee` Module.

jonathanknowles:
## Related Issue

#18

## Summary

This PR makes general improvements to the Haddock documentation for:

- The `Fee` module.
- The `FeeAdjustmentError` type.
- The `adjustForFee` function.

### DIFF
--- a/src/library/Cardano/CoinSelection/Fee.hs
+++ b/src/library/Cardano/CoinSelection/Fee.hs
@@ -132,6 +132,9 @@ data FeeOptions i o = FeeOptions
         :: DustThreshold
     } deriving Generic
 
+-- | Represents the set of possible failures that can occur when adjusting a
+--   'CoinSelection' with the 'adjustForFee' function.
+--
 newtype FeeAdjustmentError
     = CannotCoverFee Fee
     -- ^ Indicates that the given map of additional inputs was exhausted while
--- a/src/library/Cardano/CoinSelection/Fee.hs
+++ b/src/library/Cardano/CoinSelection/Fee.hs
@@ -137,8 +137,8 @@ newtype FeeAdjustmentError
     -- ^ Indicates that the given map of additional inputs was exhausted while
     --   attempting to select extra inputs to cover the required fee.
     --
-    -- Records the shortfall between the required fee and the total value
-    -- of currently selected inputs.
+    -- Records the shortfall (__/f/__ − __/s/__) between the required fee
+    -- __/f/__ and the total value __/s/__ of currently-selected inputs.
     --
     deriving (Show, Eq)
 
--- a/src/library/Cardano/CoinSelection/Fee.hs
+++ b/src/library/Cardano/CoinSelection/Fee.hs
@@ -23,16 +23,18 @@ module Cardano.CoinSelection.Fee
       -- * Fundamental Types
       Fee (..)
     , FeeEstimator (..)
-    , DustThreshold (..)
 
       -- * Fee Adjustment
     , adjustForFee
     , FeeOptions (..)
     , FeeAdjustmentError (..)
 
+      -- * Dust Processing
+    , DustThreshold (..)
+    , coalesceDust
+
       -- # Internal Functions
     , calculateFee
-    , coalesceDust
     , distributeFee
     , reduceChangeOutputs
     , splitCoin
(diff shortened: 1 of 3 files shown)


src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -131,35 +131,73 @@ data FeeOptions i o = FeeOptions
...
+--
+--  2. The __'inputs'__ set can be /augmented/, by selecting additional inputs
+--     from the specified 'CoinMap' argument.
+--
+-- == Dealing with Dust Values
+--
+-- If, at any point, a change value is generated that is less than or equal
+-- to the 'DustThreshold', this function will eliminate that change value
+-- from the 'change' set, redistributing the eliminated value over the remaining
+-- change values, ensuring that the total value of all 'change' is preserved.
```
jonathanknowles, line 190: Perhaps it would be worth referencing the 'coalesceDust' function here, as the documentation comment for that function provides a good description of how dust elimination works.

jonathanknowles: > Perhaps it would be worth referencing the `coalesceDust` function here, as the documentation comment for that function provides a good description of how dust elimination works.

Fixed in 8108efb684a937d3dfe3d855b3d327f7a439d466.

src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -131,35 +133,75 @@ data FeeOptions i o = FeeOptions
...
 newtype FeeAdjustmentError
-    = ErrCannotCoverFee Fee
-    -- ^ UTxO exhausted during fee covering
-    -- We record what amount missed to cover the fee
+    = CannotCoverFee Fee
+    -- ^ Indicates that the given map of additional inputs was exhausted while
+    --   attempting to select extra inputs to cover the required fee.
+    --
+    -- Records the shortfall between the required fee and the total value
+    -- of currently selected inputs.
```
jonathanknowles, line 141: This shortfall could be expressed more precisely, as an equation.

src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -131,35 +133,75 @@ data FeeOptions i o = FeeOptions
...
+--
+-- >>> sumInputs c = sumOutputs c + sumChange c
+--
+-- In order to pay for a transaction fee, the above equality must be
+-- transformed into an /inequality/:
+--
+-- >>> sumInputs c > sumOutputs c + sumChange c
+--
+-- The difference between these two sides represents value to be paid /by the/
+-- /originator/ of the transaction, in the form of a fee:
```
paweljakubas, line 163: I remember in the days of `cardano-sl` that there was a distinction between situtations that sender-paying-fee and recepient-paying-fee. Probably it is not the case anymore, and sender-paying-fee is tantamount to originator. But in past days it was not true.. Maybe it would be worth double check on this

src/library/Cardano/CoinSelection/Fee.hs
jonathanknowles: > This shortfall could be expressed more precisely, as an equation.

Fixed in bcf7a2e2d60f44e4d1f9516bacd35deefd63371e.

src/library/Cardano/CoinSelection/Fee.hs
jonathanknowles: I'm pretty sure that sender-pays-fee is the same as originator-pays-fee, just phrased differently.

Perhaps we could rename "originator" to "sender" to make this clearer?

src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -131,35 +133,75 @@ data FeeOptions i o = FeeOptions
     } deriving Generic
 
 newtype FeeAdjustmentError
```
jonathanknowles, line 135: This type is actually hidden from the Haddock documentation. We should give it a description so that it will be exposed.

jonathanknowles: > This type is actually hidden from the Haddock documentation. We should give it a description so that it will be exposed.

Fixed in 34ede2fd3bc75ab0a33986b032ee575f271e0126.

src/library/Cardano/CoinSelection/Fee.hs
jonathanknowles: After a brief internal discussion, we decided to leave the description as it is. :+1:

PR #61, opened 2020-04-23
re-arrange source files to better reflect module relationships

KtorZ:
result:

  ```
  src/
  ├── internal
  │   ├── Internal
  │   │   ├── Coin.hs
  │   │   ├── Invariant.hs
  │   │   └── Rounding.hs
  │   └── Internal.hs
  ├── library
  │   └── Cardano
  │       ├── CoinSelection
  │       │   └── Algorithm
  │       │       ├── LargestFirst.hs
  │       │       ├── Migration.hs
  │       │       └── RandomImprove.hs
  │       ├── CoinSelection.hs
  │       └── Fee.hs
  └── test
      ├── Cardano
      │   ├── CoinSelection
      │   │   ├── Algorithm
      │   │   │   ├── LargestFirstSpec.hs
      │   │   │   ├── MigrationSpec.hs
      │   │   │   └── RandomImproveSpec.hs
      │   │   └── TypesSpec.hs
      │   ├── CoinSelectionSpec.hs
      │   ├── FeeSpec.hs
      │   └── Test
      │       └── Utilities.hs
      ├── Internal
      │   └── CoinSpec.hs
      ├── Spec.hs
      └── Test
          └── Vector
              ├── Shuffle.hs
              └── ShuffleSpec.hs
  ```

### DIFF
--- a/src/library/Cardano/CoinSelection/Algorithm/LargestFirst.hs
+++ b/src/library/Cardano/CoinSelection/Algorithm/LargestFirst.hs
@@ -8,7 +8,7 @@
 -- This module contains an implementation of the __Largest-First__ coin
 -- selection algorithm.
 --
-module Cardano.CoinSelection.LargestFirst (
+module Cardano.CoinSelection.Algorithm.LargestFirst (
     largestFirst
   ) where
 
--- a/src/library/Cardano/CoinSelection/Algorithm/Migration.hs
+++ b/src/library/Cardano/CoinSelection/Algorithm/Migration.hs
@@ -34,7 +34,7 @@
 -- for the previous ones to be inserted before making new ones (we also say
 -- that a wallet is not "fragmented enough").
 
-module Cardano.CoinSelection.Migration
+module Cardano.CoinSelection.Algorithm.Migration
     (
       -- * Coin Selection for Migration
       selectCoins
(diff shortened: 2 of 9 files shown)


jonathanknowles:
Hi @KtorZ 

I understand the motivation to insert `Algorithm` (to differentiate algorithms from things that are not algorithms), but I'm not sure that makes sense **if we also** move `Fee` out of the `CoinSelection` namespace.

When we discussed it yesterday, I had imagined something like this:

* `Cardano.CoinSelection`
* `Cardano.CoinSelection.Algorithm.{LargestFirst, RandomImprove, Migration}`
* `Cardano.CoinSelection.Fee`

To my mind, this seems more straightforward, as:

* Placing everything in the `Cardano.CoinSelection` namespace is consistent with the name of the library.
* Library users may end up importing several Cardano-related libraries.
* Arranging for module namespaces to be consistent with library names makes it easier to guess where a particular module comes from.
* The primary purpose of the `Fee` module is actually to adjust `CoinSelection` values.

Anyway, just my thoughts.

KtorZ:
I just went with the suggestion we agreed on yesterday, which was pretty unambiguous I think. Perhaps the `Fee` part didn't caught your attention. I have no strong opinion about it being under `Cardano.CoinSelection.Fee` or not. Feel free to change.

jonathanknowles:
> I just went with the suggestion we agreed on yesterday, which was pretty unambiguous I think. Perhaps the Fee part didn't caught your attention.

You're right. I had misread that. My apologies for the confusion!

> I have no strong opinion about it being under Cardano.CoinSelection.Fee or not. Feel free to change.

Okay! I'll create a PR just to move `Fee` back into `CoinSelection` (where it will still be at a different level from the coin selection algorithms).

Issue #57, opened 2020-04-23
Port `remainingFee` calculation fix from `cardano-wallet`.


jonathanknowles:
## Background

[PR 1578](https://github.com/input-output-hk/cardano-wallet/pull/1578) for `cardano-wallet` fixed an issue with the `remainingFee` calculation.

## Decision

We should port this fix over to the `cardano-coin-selection` library _before_ reconsuming it from `cardano-wallet`.

## Related

An earlier attempt to produce a property for `remainingFee`: https://github.com/input-output-hk/cardano-coin-selection/pull/36

PR #62, opened 2020-04-23
Rework Fee Balancing  (see also input-output-hk/cardano-wallet#1578)

KtorZ:
## Related Issue

https://github.com/input-output-hk/cardano-coin-selection/issues/57

## Summary
- 0b298b9376183f1985975b38c07f7bd64c77dca5
  :round_pushpin: **rework fee balancing for coin selection**
  The previous implementation had some known issues where the algorithm would fail
to balance transactions with dangling output (i.e. small outputs that cost too
much to add). This would unfortunately trigger the no-so-invariant invariant.
This new approach completely get rid of the invariant, but introduces also some
recursion in the balancing itself. The next commit shows that the algorithm does
however converge for any coin selection, tested on 1M+ selections.

- 8141517f5af2bf8deee1853973dc784baae2ce2e
  :round_pushpin: **add a property to exercise the reduceChangeOutputs algorithm**
  
- b5c60137cd70d2b014b3dc78aa958d98b7cc4c3f
  :round_pushpin: **rename 'BalancingPolicy' to 'FeeBalancingPolicy'**
  
- 6e03ff77c98e579a5c7705b43129684ec4606dc3
  :round_pushpin: **correctly manage fee balancing in Migration steps**
  so far, testing had only been done with 'stable' fee policies where the cost of increasing
and output wouldn't have any influence on the fees. However, Byron comes with some nice fee
calculations where increasing an output may lead to an increase of fee. Therefore, the current
algorithm was in some edge cases looping indefinitely (trying to increase some change, and then
decrease it to pay for fee, and increase it back because of the surplus .. etc ..).

In practice however, Byron nodes do not require perfect balancing (whereas Jörmungandr does). So,
in Byron, the solution is still quite simple: if adding some change increases the fee beyond what
we can actually pay, we simply don't.

We still can't cope with a setup where the node would require strict balancing AND use a fee policy
that is not stable in terms of the output values. In such cases, the current algorithm would
perform pretty bad but... who on earth will really want such setup o.O ?

### DIFF
--- a/src/library/Cardano/CoinSelection/Algorithm/Migration.hs
+++ b/src/library/Cardano/CoinSelection/Algorithm/Migration.hs
@@ -1,4 +1,6 @@
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE MultiWayIf #-}
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies #-}
@@ -57,7 +59,12 @@ import Cardano.CoinSelection
     , sumInputs
     )
 import Cardano.CoinSelection.Fee
-    ( DustThreshold (..), Fee (..), FeeEstimator (..), FeeOptions (..) )
+    ( DustThreshold (..)
+    , Fee (..)
+    , FeeBalancingPolicy (..)
+    , FeeEstimator (..)
+    , FeeOptions (..)
+    )
 import Control.Monad.Trans.State
     ( State, evalState, get, put )
 import Data.List.NonEmpty
@@ -94,7 +101,7 @@ selectCoins
     -> CoinMap i
         -- ^ UTxO to deplete
     -> [CoinSelection i o]
-selectCoins feeOpts batchSize utxo =
+selectCoins FeeOptions{dustThreshold,feeEstimator,feeBalancingPolicy} batchSize utxo =
     evalState migrate (coinMapToList utxo)
   where
     migrate :: State [CoinMapEntry i] [CoinSelection i o]
@@ -120,7 +127,7 @@ selectCoins feeOpts batchSize utxo =
             in if null chgs then [threshold] else chgs
         }
       where
-        threshold = unDustThreshold $ dustThreshold feeOpts
+        threshold = unDustThreshold dustThreshold
         noDust :: Coin -> Maybe Coin
         noDust c
             | c < threshold = Nothing
@@ -148,8 +155,37 @@ selectCoins feeOpts batchSize utxo =
         -- sign of `diff` making for the right modification.
         -- We then recursively call ourselves for this might reduce the number
         -- of outputs and change the fee.
-        (c : cs) -> adjustForFee $ coinSel
-            { change = modifyFirst (c :| cs) (applyDiff diff) }
+        (c : cs) -> do
+            let coinSel' = coinSel
+                    { change = modifyFirst (c :| cs) (applyDiff diff) }
+            let costOfSurplus
+                    = fromIntegral
+                    $ C.coinToNatural
+                    $ C.distance
+                        (unFee $ estimateFee feeEstimator coinSel')
+                        (unFee $ estimateFee feeEstimator coinSel )
+            if
+                -- Adding the change costs less than not having it, so it's
+                -- worth trying.
+                | costOfSurplus < actualFee ->
+                    adjustForFee coinSel'
+
+                -- Adding the change costs more than not having it, If we don't
+                -- require strict balancing, we can leave the selection as-is.
+                | feeBalancingPolicy == RequireMinimalFee ->
+                    pure coinSel
+
+                -- Adding the change costs more than not having it. So,
+                -- depending on our balancing policy, we may stop the balancing
+                -- right here, or, if we must balance the selection discard the
+                -- whole selection: it can't be balanced with this algorithm.
+                --
+                -- Note that this last extreme case is reached when using an
+                -- unstable fee policy (where values of outputs can influence
+                -- the policy) AND, require transactions to be 100% balanced.
+                -- This is a silly thing to do.
+                | otherwise ->
+                    Nothing
       where
         applyDiff :: Integer -> Coin -> Coin
         applyDiff i c
@@ -161,10 +197,12 @@ selectCoins feeOpts batchSize utxo =
           where
             requiredFee
                 = coinToIntegral $ unFee
-                $ estimateFee (feeEstimator feeOpts) coinSel
-            actualFee
-                = coinToIntegral (sumInputs coinSel)
-                - coinToIntegral (sumChange coinSel)
+                $ estimateFee feeEstimator coinSel
+
+        actualFee :: Integer
+        actualFee
+            = coinToIntegral (sumInputs coinSel)
+            - coinToIntegral (sumChange coinSel)
 
     -- | Apply the given function to the first coin of the list. If the
     -- operation makes the 'Coin' smaller than the dust threshold, the coin is
@@ -175,7 +213,7 @@ selectCoins feeOpts batchSize utxo =
         | otherwise = c' : cs
       where
         c' = op c
-        threshold = unDustThreshold (dustThreshold feeOpts)
+        threshold = unDustThreshold dustThreshold
 
     getNextBatch :: State [a] [a]
     getNextBatch = do
--- a/src/library/Cardano/CoinSelection/Fee.hs
+++ b/src/library/Cardano/CoinSelection/Fee.hs
@@ -4,7 +4,6 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase #-}
-{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
@@ -27,6 +26,7 @@ module Cardano.CoinSelection.Fee
       -- * Fee Adjustment
     , adjustForFee
     , FeeOptions (..)
+    , BalancingPolicy (..)
     , FeeAdjustmentError (..)
 
       -- * Dust Processing
@@ -57,7 +57,7 @@ import Cardano.CoinSelection
 import Control.Monad.Trans.Class
     ( lift )
 import Control.Monad.Trans.Except
-    ( ExceptT (..), throwE )
+    ( ExceptT (..), except, throwE )
 import Control.Monad.Trans.State
     ( StateT (..), evalStateT )
 import Crypto.Random.Types
@@ -74,8 +74,6 @@ import Data.Ratio
     ( (%) )
 import GHC.Generics
     ( Generic )
-import GHC.Stack
-    ( HasCallStack )
 import Internal.Coin
     ( Coin )
 import Internal.Invariant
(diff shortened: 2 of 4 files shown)


src/library/Cardano/Fee.hs
```diff
@@ -331,26 +347,84 @@ coverRemainingFee (Fee fee) = go [] where
...
-    positiveChangeOutputs = filter (> C.zero) changeOutputs
-
-    totalChange = F.fold changeOutputs
+reduceChangeOutputs
+    :: FeeOptions i o
+    -> CoinSelection i o
+    -> Either (FeeAdjustmentError i o) (CoinSelection i o, Fee)
+reduceChangeOutputs opts s = do
+    -- The original requested fee amount
+    let Fee φ_original = estimateFee (feeEstimator opts) s
```
jonathanknowles, line 356: Hi @KtorZ 

Perhaps something like `feeOriginal` would be clearer here?

Reasons:
- For people reading the code:
    - with the prefix "fee", the comment explaining that this is a fee becomes unnecessary. 
    - it's perhaps not obvious to all that the "φ" consonant maps to "f" .
- For maintainers:
    - Latin letters are (arguably) much easier to type.

src/library/Cardano/Fee.hs
```diff
@@ -331,26 +347,84 @@ coverRemainingFee (Fee fee) = go [] where
...
+                      & coalesceDust (dustThreshold opts)
+            reduceChangeOutputs opts (s { change = chgs' })
+
+        -- the selection is leaving more fee than necessary. This typically
+        -- occurs when, after reducing a an output to pay for existing fees,
+        -- fees turn out to be less expensive than originally. In which, the
+        -- behavior will depend on whether or not the node allows transaction to
+        -- be unbalanced or not.
+        Just δ_original | δ_original > φ_original -> do
+            let Just extraChng = δ_original `C.sub` φ_original -- cf above guard
```
jonathanknowles, line 383: I notice there are quite a few partial pattern matches. (I can understand why this has happened, as it's very close to the style used in `cardano-wallet` for the equivalent piece of code.)

To avoid a partial pattern match, we could use the `C.distance` function here:
```suggestion
        Just δ_original | δ_original > φ_original -> do
            let extraChng = δ_original `C.distance` φ_original -- cf above guard
```

However, we could just as well use ordinary pattern matching here:
```hs
Just δ_original ->
    case δ_original `C.sub` φ_original of
        Just extraChange ->
            ...
        Nothing ->
            ...
```
**Advantages**:
- shorter code
- ordinary pattern matching is arguably more resistant to regressions (if a future maintainer removes the guard, the compiler won't notice that the partial pattern match is now unguarded).
- we can remove the comment

src/library/Cardano/Fee.hs
```diff
@@ -331,26 +347,84 @@ coverRemainingFee (Fee fee) = go [] where
...
+    -- The initial amount left for fee (i.e. inputs - outputs)
+    let mδ_original = sumInputs s `C.sub` (sumOutputs s `C.add` sumChange s)
+    case mδ_original of
+        -- selection is now balanced, nothing to do.
+        Just δ_original | φ_original == δ_original -> do
+            pure (s, Fee C.zero)
+
+        -- some fee left to pay, but we've depleted all change outputs
+        Just δ_original | φ_original > δ_original && null (change s) -> do
+            let Just remainder = φ_original `C.sub` δ_original -- cf above guard
```
jonathanknowles, line 366: Similar to: https://github.com/input-output-hk/cardano-coin-selection/pull/62#discussion_r414346686

We could use `C.distance` here.

src/library/Cardano/Fee.hs
```diff
@@ -331,26 +347,84 @@ coverRemainingFee (Fee fee) = go [] where
...
+            pure (s, Fee remainder)
+
+        -- some fee left to pay, and we've haven't depleted all change yet
+        Just δ_original | φ_original > δ_original && not (null (change s)) -> do
+            let Just remainder = φ_original `C.sub` δ_original -- cf above guard
```
jonathanknowles, line 371: We could use `C.distance` here. See https://github.com/input-output-hk/cardano-coin-selection/pull/62#discussion_r414346686

src/library/Cardano/Fee.hs
```diff
@@ -331,26 +347,84 @@ coverRemainingFee (Fee fee) = go [] where
...
+                    case onDanglingChange opts of
+                        SaveMoney ->
+                            pure (s, Fee C.zero)
+                        PayAndBalance ->
+                            pure (sDangling, Fee remainder)
+
+                -- If however, adding the dangling change doesn't cost more than
+                -- not having it, we might as well add it to get the money and
+                -- continue balancing!
+                _otherwise ->
```
jonathanknowles, line 420: I quite like this, as it's more visually distinct than just `_`. :+1:

src/library/Cardano/Fee.hs
```diff
@@ -331,26 +347,84 @@ coverRemainingFee (Fee fee) = go [] where
...
+                    reduceChangeOutputs opts sDangling
+
+        -- The only way to end-up here is if the user has provided an invalid
+        -- selection where outputs are trying to spend more than inputs. This is
+        -- simply forbidden.
+        _Nothing ->
```
jonathanknowles, line 426:
```suggestion
        Nothing ->
```

src/test/Cardano/FeeSpec.hs
```diff
@@ -391,8 +395,10 @@ spec = do
     describe "reduceChangeOutputs" $ do
         it "data coverage is adequate"
             (checkCoverage propReduceChangeOutputsDataCoverage)
-        it "preserves sum"
-            (checkCoverage propReduceChangeOutputsPreservesSum)
+        it "the fee balancing algorithm converges for any coin selection"
+            ( withMaxSuccess 100000
```
jonathanknowles, line 399:
```suggestion
            ( withMaxSuccess 100_000
```
(We already have `NumericUnderscores` in this module.)

src/test/Cardano/FeeSpec.hs
```diff
@@ -1013,20 +1046,47 @@ instance Arbitrary (FeeParameters i o) where
...
+                (length (inputs s) + length (outputs s) + size (change s)))
+  where
+    size :: [Coin] -> Int
+    size = sum . fmap sizeOfOne
+      where
+        sizeOfOne :: Coin -> Int
+        sizeOfOne coin
+            | coin < C.coinFromNatural 1000   = 1
+            | coin < C.coinFromNatural 10000  = 2
+            | coin < C.coinFromNatural 100000 = 3
```
jonathanknowles, line 1078: Just a nit-pick:

```suggestion
            | coin < C.coinFromNatural   1_000 = 1
            | coin < C.coinFromNatural  10_000 = 2
            | coin < C.coinFromNatural 100_000 = 3
```
My eyes must be getting old. :smile_cat:

src/library/Cardano/Fee.hs
```diff
@@ -128,21 +126,47 @@ newtype FeeEstimator i o = FeeEstimator
...
+-- can happen when a change output is small, but adding it generate more fees
+-- than not having it.
+--
+-- In case where nodes accept slightly unbalanced transactions, we may choose to
+-- save money and keep the transaction slightly unbalanced. In case where node
+-- demands exactly balanced transaction, we have no choice but to add the extra
+-- dangling change output and pay for the extra cost induced.
+data OnDanglingChange
+    = PayAndBalance
+    | SaveMoney
```
jonathanknowles, line 153: Perhaps (and this is just a suggestion) this would be easier to explain to users in the form of a **policy** regarding the _balance_:

```suggestion
data BalancePolicy
    = RequirePerfectBalance
        - ^ Generate selections that are perfectly balanced, with the
            trade-off of allowing slightly higher fees.
    | RequireMinimalFee
        - ^ Generate selections with the lowest fees possible, with the
            trade-off of allowing slightly imbalanced selections.
```

src/library/Cardano/Fee.hs
KtorZ: Actually, we need a wildcard here because of the guards and preconditions on other branch. GHC is not capable of identifying that the conditions above are mutually exclusive and that the only remaining branch actually is `Nothing`.

src/library/Cardano/Fee.hs
```diff
@@ -128,21 +126,47 @@ newtype FeeEstimator i o = FeeEstimator
...
+
+    , onDanglingChange
+        :: OnDanglingChange
+        -- ^ What do to when we encouter a dangling change output.
+        -- See 'OnDanglingChange'
     } deriving Generic
 
+-- | We call 'dangling' a change output that would be too expensive to add. This
+-- can happen when a change output is small, but adding it generate more fees
+-- than not having it.
```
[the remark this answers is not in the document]
KtorZ: `dust` refers to small UTxO, but the sense given to `small` is completely arbitrary and set by users (for instance, in the Migration algorithm we decide that anything smaller than the constant cost of transaction is dust). 

a `dangling` output however is "well-defined". It is an output that costs more to add than to not have; because adding increases the fee beyond the output value itself. In that sense, I think that you currently think as `dust` is actually `dangling` :)

src/library/Cardano/Fee.hs
```diff
@@ -128,21 +126,47 @@ newtype FeeEstimator i o = FeeEstimator
...
+--
+-- In case where nodes accept slightly unbalanced transactions, we may choose to
+-- save money and keep the transaction slightly unbalanced. In case where node
+-- demands exactly balanced transaction, we have no choice but to add the extra
+-- dangling change output and pay for the extra cost induced.
```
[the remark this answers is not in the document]
KtorZ: > Can it be stated in the form of a property involving a CoinSelection and estimateFee?

It is already!

```
            -- If fees are null and we do not require balanced selections, then
            -- the selection must leave __at least__ the fee amount.
            SaveMoney | remainder == Fee C.zero ->
                (Fee <$> delta sel') >= Just fee
```

src/library/Cardano/Fee.hs
```diff
@@ -331,26 +347,84 @@ coverRemainingFee (Fee fee) = go [] where
...
-    totalChange = F.fold changeOutputs
+reduceChangeOutputs
+    :: FeeOptions i o
+    -> CoinSelection i o
+    -> Either (FeeAdjustmentError i o) (CoinSelection i o, Fee)
+reduceChangeOutputs opts s = do
+    -- The original requested fee amount
+    let Fee φ_original = estimateFee (feeEstimator opts) s
+    -- The initial amount left for fee (i.e. inputs - outputs)
+    let mδ_original = sumInputs s `C.sub` (sumOutputs s `C.add` sumChange s)
```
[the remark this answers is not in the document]
KtorZ: I disagree here. But I understand both points :)

src/library/Cardano/Fee.hs
```diff
@@ -128,21 +126,47 @@ newtype FeeEstimator i o = FeeEstimator
...
     -- ^ Indicates that the given map of additional inputs was exhausted while
     --   attempting to select extra inputs to cover the required fee.
     --
     -- Records the shortfall (__/f/__ − __/s/__) between the required fee
     -- __/f/__ and the total value __/s/__ of currently-selected inputs.
-    --
+
+    | SpendMoreThanAvailable (CoinSelection i o)
+    -- ^ Indicates that a 'CoinSelection is malformed: the total output value is
+    -- more than what's spendable by the total input.
```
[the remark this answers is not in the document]
KtorZ: I was honestly lacking of inspiration for naming this one. I like your suggestion :+1:

src/library/Cardano/Fee.hs
jonathanknowles: > Actually, we need a wildcard here because of the guards and preconditions on other branch. GHC is not capable of identifying that the conditions above are mutually exclusive and that the only remaining branch actually is `Nothing`.

Fair enough!

src/library/Cardano/Fee.hs
KtorZ: I disagree here as the code is much easier to follow when all branches are at the same level. I consciously chose not to use nested `case .. of` here because it otherwise create a hierarchy of cases that are hard to follow.

Using guards allows for flattening it which result in a code that is much easier to follow, because each case is clearly identifiable by the structure of the code itself. Actually, the version in cardano-wallet is even more flat by not having: 

```
    let mδ_original = sumInputs s `C.sub` (sumOutputs s `C.add` sumChange s)
    case mδ_original of
```

https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Primitive/Fee.hs#L265-L301

It releis on lazyness and the fact that the `extraChng` only is evaluated when `δ_original > φ_original`.  The body of the function reads much better, although re-organizing the branches can have a dramatic effect that isn't obvious at first sight. So in the cardano-coin-selection version, I decided to have this first pattern match as a middle-ground, but still keep it flat afterwards :)

src/library/Cardano/Fee.hs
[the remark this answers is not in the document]
jonathanknowles: I agree, it's definitely subjective!

In any case, I think it's fine to leave it as is, as it doesn't affect the public API.

src/library/Cardano/Fee.hs
jonathanknowles: >  The body of the function reads much better, although re-organizing the branches can have a dramatic effect that isn't obvious at first sight. So in the cardano-coin-selection version, I decided to have this first pattern match as a middle-ground, but still keep it flat afterwards :)

Fair enough. :+1: 

In that case, I think using `C.distance` here rather than `C.sum` would be a good choice here, to avoid the partial match on `Maybe.`

src/library/Cardano/Fee.hs
```diff
@@ -128,21 +126,52 @@ newtype FeeEstimator i o = FeeEstimator
 data FeeOptions i o = FeeOptions
     { feeEstimator
         :: FeeEstimator i o
+        -- ^ Estimate fees based on selected inputs and requested outputs.
+
     , dustThreshold
         :: DustThreshold
+        -- ^ Change addresses below the given threshold will be evicted
+        -- from the created transaction. Setting 'dustThreshold' to 0
+        -- removes output equal to 0
```
jonathanknowles, line 135:
```suggestion
        -- ^ Change values that are less than or equal to this threshold will
        -- not be included in coin selections produced by the 'adjustForFee'
        -- function.
```

src/library/Cardano/Fee.hs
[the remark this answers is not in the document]
jonathanknowles: @jonathanknowles wrote:
> > Can it be stated in the form of a property involving a CoinSelection and estimateFee?
@KtorZ replied:
> It is already!

I was thinking more along the lines of adding a simple statement to the **Haddock documentation**, such as this one:

```hs
-- A selection is /balanced/ if and only if it satisfies the following
-- property:
--
-- >>> sumInputs c == sumOutputs c + sumChange c + estimateFee c
```

If a library user encounters the term "balanced" while reading through the Haddock documentation and doesn't understand what it means, it would be great if there were a definition available (without requiring them to go hunting through the source code!).

src/library/Cardano/Fee.hs
jonathanknowles: Marking this as resolved, as this has been discussed elsewhere. :+1:

src/library/Cardano/Fee.hs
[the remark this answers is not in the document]
KtorZ: Arguably, if someone doesn't know what a "balanced transaction" is, they shouldn't be using this library in the first place :wink:

src/library/Cardano/Fee.hs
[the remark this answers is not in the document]
jonathanknowles: > I was honestly lacking of inspiration for naming this one. I like your suggestion +1

Naming is hard!

src/library/Cardano/Fee.hs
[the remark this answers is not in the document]
jonathanknowles: > Arguably, if someone doesn't know what a "balanced transaction" is, they shouldn't be using this library in the first place wink

Well, I'm not so convinced about this one.

For example, the following searches don't turn up many useful results (with the search engines I tried):

* `balanced transaction` (produces a lot of junk unrelated to Blockchain)
* `"balanced transaction" blockchain` (produces a lot of definitions related to Elliptic Curves)
* `"balanced transaction" UTxO` (we're getting closer, but there are lots of articles that use it without defining it).

The Cardano documentation for "transaction" doesn't even mention the term "balanced": https://docs.cardano.org/cardano/transactions/

I suspect that our users may come with different levels of experience, and that it won't hurt to state something concretely.

src/library/Cardano/Fee.hs
```diff
@@ -128,21 +126,58 @@ newtype FeeEstimator i o = FeeEstimator
...
+-- | A /dangling/ change output is one that would be too expensive to add. This
+-- can happen when a change output is small, and including it would result in a
+-- higher fee than if it were not included.
+--
+-- In case where nodes accept slightly unbalanced transactions (i.e. when fees
+-- left are more than the minimal possible fees), we may choose to save money
+-- and keep the transaction slightly unbalanced. In case where node demands
+-- exactly balanced transactions, we have no choice but to add the dangling
+-- change output and pay for the extra cost induced.
+data BalancingPolicy
```
jonathanknowles, line 157: Given that this relates to fees, I propose the following tweak:
```suggestion
data FeeBalancingPolicy
```

PR #64, opened 2020-04-27
Documentation Improvements for `Fee` Module.

jonathanknowles:
## Related Issue

#18 

## Summary

This PR makes a selection of improvements to the documentation of the `Fee` module:

- [x] Reconciles the various descriptions of dust thresholds, so that there is a single main description, with links to that description from other places in the documentation.
- [x] Extends the explanation for `FeeEstimator` so that:
    - [x] it's clear what is required by implementers of `estimateFee`;
    - [x] it's clear why the result is an _estimate_, rather than a final fee value. 
- [x] Clarifies the explanation for `FeeBalancingPolicy`, so that:
    - [x] the definitions of _balanced_ and _unbalanced_ are clearly stated in terms of properties.
- [x] Corrects a small mistake in the description for `CoinSelectionUnderfunded`.
- [x] Corrects a few broken documentation links (to identifiers that are not directly imported).

### DIFF
--- a/src/library/Cardano/CoinSelection/Fee.hs
+++ b/src/library/Cardano/CoinSelection/Fee.hs
@@ -198,12 +198,12 @@ data FeeOptions i o = FeeOptions
 --     their value.
 --
 --   - For blockchains that do __not__ allow transactions with /unbalanced/
---     fees, specifying the 'RequirePerfectBalance' policy will always generate
+--     fees, specifying the 'RequireBalancedFee' policy will always generate
 --     selections with fees that are perfectly-balanced, even if the resulting
 --     fees are higher than could be achieved by allowing unbalanced fees.
 --
 data FeeBalancingPolicy
-    = RequirePerfectBalance
+    = RequireBalancedFee
         -- ^ Generate selections that are perfectly balanced, with the
         -- trade-off of allowing slightly higher fees.
     | RequireMinimalFee
@@ -473,7 +473,7 @@ reduceChangeOutputs opts s = do
                     case feeBalancingPolicy opts of
                         RequireMinimalFee ->
                             pure (s, Fee C.zero)
-                        RequirePerfectBalance ->
+                        RequireBalancedFee ->
                             pure (sDangling, Fee remainder)
 
                 -- If however, adding the dangling change doesn't cost more than
--- a/src/library/Cardano/CoinSelection/Fee.hs
+++ b/src/library/Cardano/CoinSelection/Fee.hs
@@ -161,15 +161,47 @@ data FeeOptions i o = FeeOptions
         -- See 'FeeBalancingPolicy'
     } deriving Generic
 
--- | A /dangling/ change output is one that would be too expensive to add. This
--- can happen when a change output is small, and including it would result in a
--- higher fee than if it were not included.
+-- | A choice of fee balancing policies for use when adjusting a coin selection.
 --
--- In case where nodes accept slightly unbalanced transactions (i.e. when fees
--- left are more than the minimal possible fees), we may choose to save money
--- and keep the transaction slightly unbalanced. In case where node demands
--- exactly balanced transactions, we have no choice but to add the dangling
--- change output and pay for the extra cost induced.
+-- == Background
+--
+-- A coin selection __'s'__ is said to have a /perfectly-balanced/ fee when it
+-- satisfies the following property:
+--
+-- >>> sumInputs s = sumOutputs s + sumChange s + estimateFee s
+--
+-- Conversely, a selection is said to have an /unbalanced/ fee when it
+-- satisfies the following property:
+--
+-- >>> sumInputs s > sumOutputs s + sumChange s + estimateFee s
+--
+-- In other words, if a coin selection has an /unbalanced/ fee, the /effective/
+-- fee is greater than the minimum fee /actually required/ by the blockchain.
+--
+-- == Balanced Fees vs Minimal Fees
+--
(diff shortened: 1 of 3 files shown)


src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -98,20 +98,41 @@ newtype Fee = Fee { unFee :: Coin }
...
 
--- | Defines the maximum size of a dust coin.
+-- | Defines the /maximum/ size of a __dust coin__.
 --
--- Change values that are less than or equal to this threshold will not be
--- included in coin selections produced by the 'adjustForFee' function.
+-- Functions that accept a 'DustThreshold' argument will generally not include
+-- values that are /less than or equal to/ this threshold in the 'change' sets
+-- of generated selections, /coalescing/ such coins together into larger
+-- coins that /exceed/ the threshold.
```
jonathanknowles, line 106:
```suggestion
-- Functions that accept a 'DustThreshold' argument will generally exclude
-- values that are /less than or equal to/ this threshold from the 'change'
-- sets of generated selections, /coalescing/ such coins together into larger
-- coins that /exceed/ the threshold.
```

src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -98,20 +98,41 @@ newtype Fee = Fee { unFee :: Coin }
...
     deriving Show via (Quiet Fee)
 
--- | Defines the maximum size of a dust coin.
+-- | Defines the /maximum/ size of a __dust coin__.
 --
--- Change values that are less than or equal to this threshold will not be
--- included in coin selections produced by the 'adjustForFee' function.
+-- Functions that accept a 'DustThreshold' argument will generally exclude
+-- values that are /less than or equal to/ this threshold from the 'change'
+-- sets of generated selections, /coalescing/ such coins together into larger
```
paweljakubas, line 105: not `change` set ?

src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -98,20 +98,41 @@ newtype Fee = Fee { unFee :: Coin }
...
+-- coins that /exceed/ the threshold.
+--
+-- Specifying a dust threshold of __/n/__ causes all coins that are less than
+-- or equal to __/n/__ to be treated as dust and coalesced together.
```
paweljakubas, line 109: you mean all coins are treated as dust and `subject to coalescence`, right?

src/library/Cardano/CoinSelection/Fee.hs
jonathanknowles: > not `change set` (rather than sets) ?

Good question. Since we are talking about "selection**s**" (plural) generated by a function (being applied many times), we must also use the plural form "set**s**" here.

src/library/Cardano/CoinSelection/Fee.hs
jonathanknowles: > you mean all coins are treated as dust and `subject to coalescence`, right?

So this specifically refers only to coins less than or equal to the threshold, rather than all coins.

If you're asking about the term "coalesced together", then I think this sounds more natural (and more direct) than coalescence (in this particular case).

src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -98,20 +98,41 @@ newtype Fee = Fee { unFee :: Coin }
...
+-- The fee estimate should be a function of the __current__ memberships of the
+-- 'inputs', 'outputs', and 'change' sets.
 --
--- The fee estimate can depend on the numbers of inputs, outputs, and change
--- outputs within the coin selection, as well as their magnitudes.
+-- Depending on the rules of the blockchain under consideration, the fee
+-- estimate may take either (or both) of the following factors into account:
+--
+--   - the number of entries in each set;
+--   - the coin value of each set member.
```
paweljakubas, line 131: when I was thinking about it one day I was also thinking that `relative coin values could also be taken into account` or `distribution of coin values` ...

jonathanknowles: > when I was thinking about it one day I was also thinking that `relative coin values could also be taken into account` or `distribution of coin values` ...

These could also be factors! (Since in principle, this library could be used with any blockchain.)

src/library/Cardano/CoinSelection/Fee.hs
```diff
@@ -130,37 +151,64 @@ data FeeOptions i o = FeeOptions
...
+-- == Balanced Fees vs Minimal Fees
+--
+-- Some blockchains /require /that fees are always /perfectly-balanced/.
+--
+-- However, for blockchains that allow /unbalanced/ fees, it is sometimes
+-- possible to /save money/ by generating a coin selection with an unbalanced
+-- fee. This may seem counterintuitive at first, but consider an individual
+-- change ouput __/c/__ of value __/v/__. If the /marginal fee/ __/f/__
+-- associated with __/c/__ is greater than its value __/v/__, then we will
+-- /save money/ by __not__ including __/c/__ within 'change'.
```
paweljakubas, line 190: would be perfect to see it in real examples. show fee balancing policy that exemplifies it and see in code here

jonathanknowles: > would be perfect to see it in real examples. show fee balancing policy that exemplifies it and see in code here. Without it it is still close to perfect;-)

Great idea. I'll see if I can include it in another PR!

Issue #19, opened 2020-02-26
Release `cardano-coin-selection` library to Hackage and Stackage.


jonathanknowles:
# Context

As part of the work to create a separate [**coin selection and fee balancing library**](https://github.com/input-output-hk/cardano-wallet/milestone/58), we'd like for the library to be consumable by the wider Haskell community.

# Decision

Release the library to both [Hackage](https://hackage.haskell.org/) and [Stackage](https://www.stackage.org/).

# Prerequisites

The following issues **must** be closed before we can make a release:

From `cardano-wallet`:

<a href="https://github.com/input-output-hk/cardano-wallet/issues/1366"><img src="https://img.shields.io/github/issues/detail/state/input-output-hk/cardano-wallet/1366?style=for-the-badge" /></a>
<a href="https://github.com/input-output-hk/cardano-wallet/issues/1379"><img src="https://img.shields.io/github/issues/detail/state/input-output-hk/cardano-wallet/1379?style=for-the-badge" /></a>
<a href="https://github.com/input-output-hk/cardano-wallet/issues/1380"><img src="https://img.shields.io/github/issues/detail/state/input-output-hk/cardano-wallet/1380?style=for-the-badge" /></a>

From `cardano-coin-selection`:

<a href="https://github.com/input-output-hk/cardano-coin-selection/issues/17"><img src="https://img.shields.io/github/issues/detail/state/input-output-hk/cardano-coin-selection/17?style=for-the-badge" /></a>
<a href="https://github.com/input-output-hk/cardano-coin-selection/issues/18"><img src="https://img.shields.io/github/issues/detail/state/input-output-hk/cardano-coin-selection/18?style=for-the-badge" /></a>
<a href="https://github.com/input-output-hk/cardano-coin-selection/issues/30"><img src="https://img.shields.io/github/issues/detail/state/input-output-hk/cardano-coin-selection/30?style=for-the-badge" /></a>

# Acceptance Criteria

- [x] The [`cardano-coin-selection` library](https://github.com/input-output-hk/cardano-coin-selection/) _must_ be available on [Hackage](https://hackage.haskell.org/), _including_ Haddock documentation.
- [x] The [`cardano-coin-selection` library](https://github.com/input-output-hk/cardano-coin-selection/) _must_ be available on [Stackage](https://www.stackage.org/), _including_ Haddock documentation.

# Development 

[...]

PR #67, opened 2020-04-29
Prepare for v1.0.0 release.

jonathanknowles:
## Related Issue

#19 

## Summary:

This PR:

- [x] Removes redundant dependencies from `package.yaml`.
- [x] Updates the change log with the correct release date.
- [x] The "Added", "Changed", and "Removed" headings do not apply to the initial release, and so have been removed from the change log. (These sections will be added in subsequent releases.)
- [x] Updates `stack.yaml` to the most recent LTS.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,13 +1,3 @@
-## [1.0.0] - 2020-04-23 
+## [1.0.0] - 2020-04-29
 
-### Added
-
-- TODO
-
-### Changed
-
-N/A
-
-### Removed
-
-N/A
+Initial release.
--- a/package.yaml
+++ b/package.yaml
@@ -43,7 +43,6 @@ library:
   - cryptonite
   - deepseq
   - quiet
-  - quiet
   - text
   - transformers
 
--- a/package.yaml
+++ b/package.yaml
@@ -64,6 +64,7 @@ tests:
       ghc-options:
       - -Werror
     dependencies:
+    - QuickCheck
     - bytestring
     - cardano-coin-selection
     - containers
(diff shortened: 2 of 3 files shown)


jonathanknowles:
@rvl wrote:
> It will probably cause less trouble if we keep libs that will be used by cardano-wallet on the same Stackage LTS series as cardano-wallet.

That's a good point.

As `cardano-wallet` currently uses `lts-14.25`, we could use this version. However, any change we make is likely to become out of date very quickly.

We could change this entry to link to https://raw.githubusercontent.com/input-output-hk/cardano-haskell/master/snapshots/cardano-1.11.0.yaml. Can you see any disadvantages to this approach?

rvl:
Main problem with LTS 15.x is that it's using GHC 8.8 when cardano-wallet is on GHC 8.6.

I think it's enough to keep on the same LTS major version (14.x) as cardano-wallet. For this package, the cardano-haskell snapshot has too many irrelevant packages.

jonathanknowles:
@rvl wrote:
> Main problem with LTS 15.x is that it's using GHC 8.8 when cardano-wallet is on GHC 8.6.
> 
> I think it's enough to keep on the same LTS major version (14.x) as cardano-wallet. For this package, the cardano-haskell snapshot has too many irrelevant packages.

Okay! I've changed it to use `lts-14.25`. (I did experiment with using the same resolver reference as `cardano-wallet`, but that caused stack to start pulling in dozens of extra unnecessary packages, as predicted.)

Fixed in d4374a225ad77366735998e8115216273172d335.

Issue #39, opened 2020-04-09
The Largest-First algorithm should pay for outputs collectively.


jonathanknowles:
## Context

The **Largest-First** algorithm is used as a fall-back for cases when more sophisticated algorithms fail to produce a coin selection.

However, our current implementation is **broken** in a subtle way. Rather than considering the total collective value of all outputs when attempting to find a selection of inputs (as it should), our implementation instead considers each output independently, and for each output, attempts to find a selection of inputs to cover _just that output_.

This leads to an unnecessary **cardinality restriction**, namely that a single unique input can only be used to pay for at most one unique output, which means that any set of **_n_** outputs can only be paid for if the UTxO has at least **_n_** inputs. Under this restriction, the algorithm can fail even if the total value of inputs is sufficient to cover the required outputs.

## Suggested Fix

We should adjust our current implementation so that the algorithm considers the total value of all outputs collectively, rather than considering them individually.

Under this relaxed scheme, it will always be possible to pay for a given set of outputs of total value **_v_** if the total value **_u_** of the UTxO satisfies **_u_** >= **_v_**.

PR #73, opened 2020-05-07
Make the `Largest-First` algorithm pay for outputs collectively.

jonathanknowles:
## Related Issue

#39 

## Summary

This PR changes the reference implementation of the `Largest-First` coin selection algorithm to match the [specification](https://github.com/jonathanknowles/CIPs/blob/CIP-JonathanKnowles-CoinSelectionAlgorithms/CIP-JonathanKnowles-CoinSelectionAlgorithms/CIP-JonathanKnowles-CoinSelectionAlgorithms.md#largest-first).

After applying this PR, the algorithm should always be successful at paying for a given set of outputs of total value **_v_** provided that the total value **_u_** of available inputs satisfies **_u_** ≥ **_v_**. (Unless the maximum input count has been exceeded.)

### Old implementation (before applying this PR)
    
Rather than considering the total collective value of all outputs when attempting to find a selection of inputs, the old implementation considered each output independently, and for each output, attempted to find a selection of inputs to cover just that output.
    
This created an unnecessary cardinality restriction, namely that a single unique input could only be used to pay for at most one unique output, which meant that any set of **_n_** outputs could only be paid for if there were at least **_n_** inputs available. Under this restriction, the algorithm could fail _even if_ the total value of inputs was sufficient to cover the required outputs.
    
### New implementation (after applying this PR)
    
The updated algorithm now considers the _total_ value of all outputs collectively, rather than considering them individually.
    
Under this relaxed scheme, it will always be possible to pay for a given set of outputs of total value **_v_** if the total value **_u_** of available inputs satisfies **_u_** ≥ **_v_**. (Unless the maximum input count has been exceeded.)

### DIFF
--- a/src/test/Cardano/CoinSelection/Algorithm/LargestFirstSpec.hs
+++ b/src/test/Cardano/CoinSelection/Algorithm/LargestFirstSpec.hs
@@ -27,7 +27,7 @@ import Cardano.CoinSelection
 import Cardano.CoinSelection.Algorithm.LargestFirst
     ( largestFirst )
 import Cardano.CoinSelectionSpec
-    ( CoinSelProp (..)
+    ( CoinSelectionData (..)
     , CoinSelectionFixture (..)
     , CoinSelectionTestResult (..)
     , coinSelectionUnitTest
@@ -214,9 +214,9 @@ spec = do
 
 propInputDecreasingOrder
     :: Ord i
-    => CoinSelProp i o
+    => CoinSelectionData i o
     -> Property
-propInputDecreasingOrder (CoinSelProp utxo txOuts) =
+propInputDecreasingOrder (CoinSelectionData utxo txOuts) =
     isRight selection ==>
         let Right (CoinSelectionResult s _) = selection in
         prop s
--- a/src/test/Cardano/CoinSelection/Algorithm/LargestFirstSpec.hs
+++ b/src/test/Cardano/CoinSelection/Algorithm/LargestFirstSpec.hs
@@ -205,9 +205,6 @@ spec = do
 
     describe "Coin selection: largest-first algorithm: properties" $ do
 
-        it "forall (UTxO, NonEmpty TxOut), there's at least as many selected \
-            \inputs as there are requested outputs"
-            (property $ propAtLeast @TxIn @Address)
         it "forall (UTxO, NonEmpty TxOut), for all selected input, there's no \
             \bigger input in the UTxO that is not already in the selected \
             \inputs"
@@ -216,21 +213,6 @@ spec = do
 --------------------------------------------------------------------------------
 -- Properties
 --------------------------------------------------------------------------------
-
-propAtLeast
-    :: Ord i
-    => CoinSelProp i o
-    -> Property
-propAtLeast (CoinSelProp utxo txOuts) =
-    isRight selection ==>
-        let Right (CoinSelectionResult s _) = selection in
-        prop s
-  where
-    prop (CoinSelection inps _ _) =
-        length inps `shouldSatisfy` (>= length txOuts)
-    selection = runIdentity $ runExceptT $ selectCoins largestFirst
-        $ CoinSelectionParameters utxo txOuts selectionLimit
-    selectionLimit = CoinSelectionLimit $ const 100
 
 propInputDecreasingOrder
     :: Ord i
--- a/src/test/Cardano/CoinSelection/Algorithm/LargestFirstSpec.hs
+++ b/src/test/Cardano/CoinSelection/Algorithm/LargestFirstSpec.hs
@@ -48,13 +48,14 @@ import Data.Function
 import Data.Functor.Identity
     ( Identity (runIdentity) )
 import Test.Hspec
-    ( Spec, describe, it, shouldSatisfy )
+    ( Spec, describe, it, shouldBe, shouldSatisfy )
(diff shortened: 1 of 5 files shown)


src/test/Cardano/CoinSelection/Algorithm/LargestFirstSpec.hs
```diff
@@ -245,3 +234,17 @@ propInputDecreasingOrder (CoinSelProp utxo txOuts) =
...
     selectionLimit = CoinSelectionLimit $ const 100
+
+--------------------------------------------------------------------------------
+-- Utilities
+--------------------------------------------------------------------------------
+
+-- Returns true if (and only if) the given error value is one that can be
+-- thrown by the Largest-First algorithm.
+--
+isValidLargestFirstError :: CoinSelectionError -> Bool
```
paweljakubas, line 245: So we have still errors present that stopped to be valid in current impl of algo, right? Why not to prune them? We can always model those `exhausted` and `not-properly fragmented-inputs` to show that with new implementation they pass, no ?

paweljakubas: Or they are going to be used in `random` (ie., those errors) ? If not, I feel they are superfluous...

jonathanknowles: @paweljakubas wrote:
>  So we have still errors present that stopped to be valid in current impl of algo, right? Why not to prune them? We can always model those exhausted and not-properly fragmented-inputs to show that with new implementation they pass, no ?

Good question!

Those errors are still used by the `Random-Improve` algorithm.

We **_could_** make it so that the algorithms return different sets of errors, but this would break backwards compatibility in the public API. (We'd need to increment the major version to do this.)

If we decide to do that, perhaps we should do it in a future PR. Thoughts?

paweljakubas: ok, if the errors are used in `random` then it seems ok to me. It even has additional  documentation aspect that the largest first does not deal with them expresses explicitely;-)

jonathanknowles:
> Q: So after collectively addressing all inputs to be covered (instead of one to one) we have more robust algo. Let's assume we want to cover `[2, 1, 2]` inputs. and have `[4, 1, 1]` utxos. In old implementation:
> `2` could be attributed `4`, but the next one `1` would be not enough to cover the next`2` and we would end up with error. Right now, `4` will cover `[2,2]` and `[1]` will cover the rest?

If we have inputs of `[1, 1, 4`] and want to pay for outputs of `[2, 1, 2]`, the algorithm will:
1. Calculate the total value of all outputs: `5`
2. Place the inputs into a list of descending order: `[4, 1, 1`].
3. Take just enough inputs from the head of the list to pay for the total value of all outputs. In this case, it will be `[4, 1]`.
4. Since we have paid for the outputs exactly, there will be no change.

If we have inputs of `[1, 2, 4`] and want to pay for outputs of `[2, 1, 2]`, the algorithm will:
1. Calculate the total value of all outputs: `5`
2. Place the inputs into a list of descending order: `[4, 2, 1`].
3. Take just enough inputs from the head of the list to pay for the total value of all outputs. In this case, it will be `[4, 2]`, with a total selected input value of `6`.
4. Since there's an excess (`6 - 5 = 1`), the algorithm generates a change value of `1`.

src/test/Cardano/CoinSelection/Algorithm/LargestFirstSpec.hs
```diff
@@ -116,118 +127,113 @@ spec = do
...
+                , utxoInputs = [12,10,17]
+                , txOutputs = [40]
                 })
 
         coinSelectionUnitTest largestFirst
-            "UTxO balance sufficient, fragmented enough, but maximum input \
-            \count exceeded"
-            (Left $ InputLimitExceeded $ InputLimitExceededError 9)
+            "UTxO balance not sufficient: case #2"
+            (Left $ InputValueInsufficient $ InputValueInsufficientError
```
paweljakubas, line 193: this one is surprise it is error. we have to cover `[12,10,17]` and maxNumOfInputs is forgiving, and we have `40` to cover all this

paweljakubas: ok. I thought it was opposite inputs/utxos. nevertheless, why not to add the case I was thinking about ;-)

jonathanknowles: > this one is surprise it is error. we have to cover `[12,10,17]` and maxNumOfInputs is forgiving, and we have `40` to cover all this

That one is actually correct, as:

* The available inputs are `[12,10,17]` (total value = `39`)
* The requested outputs are `[40,1,1,1]` (total value = `43`)

So there's not enough value in the available inputs to pay for all of the outputs.

src/test/Cardano/CoinSelection/Algorithm/LargestFirstSpec.hs
jonathanknowles: > ok, if the errors are used in `random` then it seems ok to me. It even has additional documentation aspect that the largest first does not deal with them expresses explicitly;-)

Agreed! Though I do also agree we could consider making this more explicit in the types in a future version.

PR #81, opened 2020-05-14
Add Specification Properties.

jonathanknowles:
## Related Issue

#21 

## Summary

This PR adds missing tests for the following properties, as defined by the [coin selection specification](https://github.com/input-output-hk/cardano-coin-selection/issues/21):

- Pre-conditions:
    - [x] `value inputsAvailable ≥ value outputsRequested`
- Post-conditions:
    - [x] `outputsSelected = outputsRequested`
    - [x] `inputsSelected ⊆ inputsAvailable`
    - [x] `inputsRemaining ⊆ inputsAvailable`
    - [x] `inputsSelected ⋂ inputsRemaining = ∅`
    - [x] `inputsSelected ⋃ inputsRemaining = inputsAvailable`
    - [x] `value inputsSelected = value outputsSelected + value change`

This PR also checks that the above properties are satisfied by all algorithms that implement the `CoinSelectionAlgorithm` interface.

### DIFF
--- a/src/test/Cardano/CoinSelectionSpec.hs
+++ b/src/test/Cardano/CoinSelectionSpec.hs
@@ -140,7 +140,7 @@ spec = do
     lowerConfidence = Confidence (10^(6 :: Integer)) 0.75
 
 --------------------------------------------------------------------------------
--- Properties
+-- Coin Map Properties
 --------------------------------------------------------------------------------
 
 prop_CoinMap_coverage
@@ -261,6 +261,10 @@ prop_coinMapToList_orderDeterministic u = monadicIO $ QC.run $ do
     return $
         cover 10 (list0 /= list1) "shuffled" $
         list0 == coinMapToList (coinMapFromList list1)
+
+--------------------------------------------------------------------------------
+-- Coin Selection Properties
+--------------------------------------------------------------------------------
 
 prop_coinSelection_mappendPreservesKeys
     :: (Ord i, Ord o, Show i, Show o)
--- a/src/test/Cardano/CoinSelectionSpec.hs
+++ b/src/test/Cardano/CoinSelectionSpec.hs
@@ -170,6 +170,10 @@ coinSelectionAlgorithmGeneralProperties algorithm algorithmName =
             property $
             prop_algorithm_inputsAvailable_inputsRemaining algorithm
 
+        it "inputsSelected ⋂ inputsRemaining = ∅" $
+            property $
+            prop_algorithm_inputsSelected_inputsRemaining algorithm
+
 --------------------------------------------------------------------------------
 -- Coin Map Properties
 --------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


src/test/Cardano/CoinSelectionSpec.hs
```diff
@@ -139,8 +145,35 @@ spec = do
     lowerConfidence :: Confidence
     lowerConfidence = Confidence (10^(6 :: Integer)) 0.75
 
+-- A collection of general properties that should apply to all algorithms
+-- that implement the 'CoinSelectionAlgorithm' interface.
+--
+coinSelectionAlgorithmGeneralProperties
```
paweljakubas, line 151: why not to use `Common` or `Generic` ? it sounds better to me (I know you know better as you are native speaker :-) )

src/test/Cardano/CoinSelectionSpec.hs
```diff
@@ -304,6 +341,107 @@ prop_CoinSelectionData_coverage (CoinSelectionData inps outs) = property
...
+        $ CoinSelectionLimit
+        $ const $ fromIntegral $ F.length csdInputsAvailable
+
+isSubmapOf :: Ord k => CoinMap k -> CoinMap k -> Bool
+isSubmapOf (CoinMap a) (CoinMap b) = a `Map.isSubmapOf` b
+
+intersection :: Ord k => CoinMap k -> CoinMap k -> CoinMap k
+intersection (CoinMap a) (CoinMap b) = CoinMap $ a `Map.intersection` b
+
+union :: Ord k => CoinMap k -> CoinMap k -> CoinMap k
```
paweljakubas, line 442: is this really needed when we have https://github.com/input-output-hk/cardano-coin-selection/blob/master/src/library/Cardano/CoinSelection.hs#L123 ?

paweljakubas: ah, ok. you wanted to make it explicit

src/test/Cardano/CoinSelectionSpec.hs
jonathanknowles: > why not to use Common or Generic ? it sounds better to me (I know you know better as you are native speaker :-) )

Good question!

I'm trying to capture a contrast between _specialized_ and _generalized_.

We already have tests for properties that are specific (or specialized) to particular algorithms. For example, for `Largest-First`, we have a specialized property test that checks that all selected inputs are greater than or equal (in coin value) to the remaining inputs not selected.

However, the properties in this PR are all general (non-specific), as they apply to all implementations.

We could also use the term "common" here. I tried to avoid the word "generic" as it could be confused with the concept of generic programming.

src/test/Cardano/CoinSelectionSpec.hs
jonathanknowles: > is this really needed when we have https://github.com/input-output-hk/cardano-coin-selection/blob/master/src/library/Cardano/CoinSelection.hs#L123 ?

Strictly speaking it's not needed!

Here my motivation was to make the tests more readable. Defining `union` (arguably) makes the test more readable at the call site.

In particular, the following two lines are now directly comparable to one another:

```hs
(selected `intersection` remaining) `shouldBe` mempty
(selected `union`        remaining) `shouldBe` available
```

Whereas if we just use `<>` instead of `union`, it's no longer immediately clear that we're invoking a union operation (IMO).

Issue #34, opened 2020-04-08
Revise test suite.


jonathanknowles:
## Context

This issue collects together tasks relating to simplifying the test suite for `cardano-coin-selection`.

## Tasks

- [ ] **`TxIn` type** (used for testing)

    This could probably be simplified, as we test code only relies on the fact that generated values are unique and can be ordered.

    Additionally, we could replace all the various `Arbitrary` instances (which are mostly alike) with a single instance.

- [ ] **`Arbitrary` instances**

    Many of these are duplicates of one another, or very similar. We should consider whether or not it would be worth merging similar instances.

- [ ] **`TODO` comments**

    These should either be processed or deleted.

- [ ] **`NOTE` comments**

    These should either be processed or deleted.

PR #82, opened 2020-05-15
Retire wallet-specific types from the test suite.

jonathanknowles:
## Related Issue

#34 

## Summary

This PR retires the following wallet-specific types from the test suite:

- [x] `TxIn`
- [x] `TxOut`
- [x] `Address`
- [x] `Hash`

The following simpler types are introduced as replacements:

- [x] `InputId`
- [x] `OutputId`

Both of the above types are based on a common `UniqueId` type, which just wraps a `ByteString`.

## Motivation

Before this change, QuickCheck failures would often produce output similar to the following:
    
```hs
TxIn {txinId = Hash "\GS\STX.\t\CAN\STX*/\DC3'@@7$;)4\STX\DC2*@+\RS0/=-\SUB%\b\RS?", txinIx = 1}
```

As a single value, this is already pretty difficult to read. Test failures involving multiple such values can be almost unreadable!

By contrast, `Show` instances for `InputId` and `Output` produce output similar to:

```hs
InputId 1a2b3c4d
```

### DIFF
--- a/src/test/Cardano/Test/Utilities.hs
+++ b/src/test/Cardano/Test/Utilities.hs
@@ -4,9 +4,9 @@
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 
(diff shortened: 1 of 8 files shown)


src/test/Cardano/Test/Utilities.hs
```diff
@@ -54,67 +54,86 @@ import Cardano.CoinSelection.Fee
...
 import qualified Internal.Coin as C
 
 --------------------------------------------------------------------------------
--- Addresses
+-- Unique Identifiers
 --------------------------------------------------------------------------------
 
-newtype Address = Address
-    { unAddress :: ByteString }
+newtype UniqueId (tag :: Symbol) = UniqueId { unUniqueId :: ByteString }
```
paweljakubas, line 93: :100:

jonathanknowles:
> OK, so if the library has introduced polymorphic inputs/outputs, then it's possible to run unit tests with simpler types, so long as they behave the same.

To give a little background:

We introduced polymorphic identifier types to the library some time ago, as coin selection algorithms don't really care about the specifics of addresses and transaction identifiers. (They generally only care that values of these types can be _ordered_ and tested for equality.)

